Issue #116, opened 2026-02-26
Support non-filtered `predict_times` for trajectory-based evaluation


mattlevine22:
- To start, I suggest we support the case where all forecast_times > all obs_times
- do `sample(..., DynamicalModel, obs_times, obs_values, forecast_times)`
- for now, interpret using the inference method (filters can do the forecasting, and unrollers can keep unrolling)
- CD_Dynamax cdnlgssm and cdlgssm (not cdnlssm) offer `filter_and_forecast`

PR #119, opened 2026-02-26
Adding `predict_times` option

mattlevine22:
Addressing #116 .

- leverage `filter_and_forecast` in CD-Dynamax
- auto-continue forecasts in case of Simulators

In the future, we could nest `with Simulator -> with Filter`, but this seemed to be rather complicated at first cut.

- Current version needs another pass to simplify / de-AI-ify a bit; making a draft PR to make this easier.


DanWaxman:
Closing this in favor of #135.

PR #121, opened 2026-03-02
Include PFJax Support

DanWaxman:


pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "dynestyx"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = ">=3.12,<3.14"
```
mattlevine22: Was it PFJax or blackjax that created the need for <3.14? Let's try to remember this somehow so we aren't struggling when we need to update.

mattlevine22 (commented):
Overall looks good to me, would be happy with including with a commented notebook (or dropping the notebook as a private thing)

pyproject.toml
DanWaxman: As commented above, it's actually CD-Dynamax

docs/deep_dives/particle_hmc.ipynb
```diff
@@ -0,0 +1,596 @@
+{
```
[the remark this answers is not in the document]
mattlevine22: Sounds great. Do we want it to be a "deep dive" in that case?

We may want a place for non-doc scripts/notebooks for when we do something quickly and don't want to "polish" it. Alternatively we can banish this practice and push ourselves to have nice notebooks (we've been pretty good about this so far).

DanWaxman:
I'm going to close this, it's super duper stale by now. Though still hope to add PFJax eventually :)

PR #124, opened 2026-03-02
Add GitHub Actions workflow for Python package upload

eb8680:
This file is copied from https://github.com/BasisResearch/effectful/blob/master/.github/workflows/python-publish.yml


DanWaxman (approved):
Thanks Eli!!

PR #126, opened 2026-03-03
Add Cuthbert KF & Update Cuthbert

DanWaxman:
This adds the cuthbert KF, and upgrades to a newer version of cuthbert. This version is a fork due to two issues: (i) some instability in gradient estimates from the Kalman filter, and (ii) a change in cuthbert about observations at $t_0$. I have opened a issue/PR about (i), and asked about cuthbert's long-term intentions for (ii). We'll need some more elaborate re-indexing for (ii) depending on their answers.

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -23,7 +23,6 @@ from dynestyx.inference.filter_configs import (
     ContinuousTimeKFConfig,
     ContinuousTimeUKFConfig,
     EKFConfig,
-    EnKFConfig,
     KFConfig,
     PFConfig,
     UKFConfig,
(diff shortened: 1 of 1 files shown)


DanWaxman:
Added stop-gradient, and updated tests and a deep dive notebook a bit.

tests/fixtures.py
```diff
@@ -23,6 +23,7 @@ def _n_particles(base: int) -> int:
     ContinuousTimeKFConfig,
     ContinuousTimeUKFConfig,
     EKFConfig,
+    EnKFConfig,
```
mattlevine22: spurious EnKF

Issue #50, opened 2026-02-01
Forecast evaluations will need a posterior "rollout" mode


mattlevine22:
We will want to make notebooks [like this in cd-dynamax](https://github.com/hd-UQ/cd_dynamax/blob/dev-numpyro-api/demos/numpyro/notebooks/lorenz63_nndrift_sgd_fit_to_data_tutorial_newAPI.ipynb).

More generally, we will want to evaluate rollout forecasts coming from our posterior.

However, we don't have the infrastructure to do this. 
- Should think carefully about how.
- Should this be determined by an additional "times" field in the Condition(Context(...))?

PR #135, opened 2026-03-10
Predict Times

DanWaxman:
Addresses #50 #116 

## Summary

This PR introduces first-class support for `predict_times`-driven simulation and filter rollout across simulators and filtering integrations, with consistent output sites, shape handling, and docs/tests coverage.

- Adds/expands simulator + filter composition so models can:
  - condition on `obs_times`/`obs_values`
  - optionally roll out trajectories at `predict_times`
  - expose deterministic rollout outputs (`f_predicted_times`, `f_predicted_states`, `f_predicted_observations`) in a consistent way.
- Refactors simulator internals (notably `BaseSimulator`/`DiscreteTimeSimulator`/SDE/ODE flow) to better separate:
  - conditioning paths vs forward simulation paths
  - n-simulation vectorized execution vs single-trajectory conditioned execution
  - filter-rollout segmentation behavior.
- Unifies particle-filter posterior conversion logic across integrations via shared utility code (`particles_to_delta_mixtures`), with standardized expected tensor shapes and cleaner backend plumbing.
- Improves control/time validation and alignment behavior (including union-of-times handling where relevant), and clarifies edge-case behavior for missing grids.
- Adds substantial test coverage for matrixed input combinations and shape contracts:
  - new input-matrix tests for `obs_times`/`obs_values`/`predict_times`
  - predictive shape tests for `num_samples` and `n_simulations`
  - updated smoke/science/model tests to reflect new rollout/output conventions.
- Updates API docs/tutorials to clearly explain:
  - when to use `obs_times` vs `predict_times`
  - required pairing of `obs_times` and `obs_values`
  - deterministic site names and resulting output shapes.
- Updates notebooks with concrete `predict_times` examples, including:
  - rollout/forecasting after filtering
  - back-fill interpolation/reconstruction between observation times.

## Motivation

Previously, prediction-time rollout and simulator/filter interoperability were harder to reason about and less uniformly tested/documented, especially around mixed `obs_times`/`predict_times` workflows and shape conventions.  
This PR makes that behavior explicit, consistent, and verifiable: users can condition and forecast in one coherent interface, while downstream analysis/tests/docs rely on stable site naming and tensor shape contracts.

## Performance

This solution entails some additional checks + building and storing filtered distributions at all `obs_times` points during a sample statement. In our experiments, this causes a ~20% overhead. This is non-negligible, but seems worth it to have the prediction feature.


mattlevine22:
@DanWaxman I **THINK** this PR is more-or-less ready to be looked at.

Issue #138, opened 2026-03-16
Add support for missing data in discrete-time models


dimkab:
Modify `DiscreteTimeSimulator::_simulate` to support `NaN`s in some observation components.
- [ ] l410: adjust the condition to take partial NaNs into account (i.e. enter the block only if all components are NOT NaNs)
- [ ] l484, l490: adjust the obs for DiracObservations
- [ ] add utilities to factor multivariate Gaussians [optional, another PR?]

Prefer no dispatches :)

Tests: TBD

DanWaxman:
Is there anything left here after #239 #234 (cc @mattlevine22)? I suppose we need to get things working with continuous-time models, but that seems more on CD-Dynamax and/or discretizer + discrete-time filter/smoother.

mattlevine22:
Yeah I wanna do DiracIdentity next...it should probably handle full dropout at least. Could open separate issue for that if you want. @dimkab how are you handling things in your custom numpyro code these days...I'd really like to support what you're doing via dynestyx settings if possible.

dimkab:
@mattlevine22 essentially as in my PR:
https://github.com/BasisResearch/collab-dynmodels/blob/55b250699d7881998a0633fb8abb6d144924b5aa/src/inference_models/likelihood_engines.py#L278

mattlevine22:
Thanks for linking @dimkab !

TLDR: 
1. Have you considered using `LinearGaussianObservation` instead? Might adequately mimic what you have already. Happy to talk about what might be different.
2. At the end, I propose some code for how to vectorize things more broadly (something dynestyx and I think your collab-dynmodels similarly struggle with).

Details...
If I understand your PR correctly, you use `numpyro.contrib.control_flow.scan` to step through transitions in the general missingness case, right? If so, I think our main Simulator w/ non-Dirac interface (e.g., create a LinearGaussianObservation with a desired noise level) does what you need. 

The main benefit of DiracIdentity stuff is to`vmap` instead of scan. But once there is missingness, numpyro really wants us to do a scan...I think your above code reflects this challenge. Seems like the one case where your code does vectorized stuff is in `skip=True` mode where all measurements were dropped in a block (not partially-missing); we didn't implement this specific idea for Simulator+Dirac, but we could if its important.

More generally (curious about @DanWaxman 's take), we could do a lot more vectorization under missingness if we let go of the PPL beauty under-the-hood. Something like:

```python

[...]

PR #140, opened 2026-03-16
support for missing data

dimkab:
A partial support for missing data is provided for :class:`DiscreteTimeSimulator` in the case that `obs_values` has entire rows of `nan`s. Since `DiscreteTimeStateEvolution` always has `t_now` and `t_next`, we can just remove the missing rows from `obs_values` and the corresponding times from `obs_times` before proceeding with the likelihood computation - which will be carried out only on observed times without losing any information.

Partially observed case will be implemented in #149.

Cf. #138


dimkab:
Superseded elsewhere

PR #149, opened 2026-03-18
Missing data: partial missingness, unroll_missing, diagonal obs models

dimkab:
## Summary

- Adds per-dimension (partial) missingness support to `DiscreteTimeSimulator`
- `unroll_missing=True` (default) produces full-length output with gap imputation when entire rows are missing
- New `ObservationModel.masked_log_prob` interface for per-dimension scoring
- New `DiagonalLinearGaussianObservation` and `DiagonalGaussianObservation` observation models
- Tutorial notebook (08) covering three missingness patterns + gap imputation + interacting particles

### `dynestyx/simulators.py`

`unroll_missing: bool = True` field on `DiscreteTimeSimulator`. The `_simulate` method dispatches to:

- `_simulate_missing_scan` — when partial NaNs or `unroll_missing=True` with missing rows
- `_simulate_row_filter` — when `unroll_missing=False` with missing rows
- `_simulate_plate` / `_simulate_scan` — no missingness

`_simulate_missing_scan` has two sub-paths:

**Sub-path A (`DiracIdentityObservation`) — factor + masked-sample:**
- Observed dims: `numpyro.factor(obs_lp)` scores the transition log-prob
- Latent dims: `numpyro.handlers.mask(mask=~obs_mask) { numpyro.sample(trans_dist) }` — the mask zeros out both model and guide ELBO contributions at observed dims, so `AutoNormal` allocates parameters but they are inert (zero gradient)
- Two variants: `_step_dirac_partial` (per-dim vector mask, requires `Independent(Normal, 1)` transitions, uses `base_dist`) and `_step_dirac_wholerow` (scalar mask, works with any transition including `MultivariateNormal`)

**Sub-path B (non-Dirac obs models):**
- State is always latent (sampled unconditionally)
- Observation scoring via `obs_model.masked_log_prob(y, obs_mask, x, ...)` for partial rows, or full `log_prob` zeroed at missing rows via `numpyro.handlers.mask`

### `dynestyx/models/`

- `ObservationModel.masked_log_prob`: scores only observed dims; default raises `NotImplementedError`
- `DiagonalLinearGaussianObservation`: linear H with diagonal R, supports `masked_log_prob`
- `DiagonalGaussianObservation`: nonlinear obs function with diagonal R, supports `masked_log_prob`

### Tests

- `tests/test_with_missing_data.py` — 39 tests covering all combinations of:
  - Models: LTI, diagonal obs (linear + nonlinear), particle SDE, interacting particles
  - Missingness: none, random, sequential, block, partial
  - `unroll_missing`: True/False
  - Inference: MCMC and SVI
- Smoke/full mode via `DYNESTYX_SMOKE_TEST` env var (set in `scripts/test.sh`)
  - Smoke: small T, few samples, gradient-flow checks only
  - Full: convergence + posterior accuracy checks

### Tutorial notebook (08)

Sections: whole-row block missingness (default + row-filter paths, side-by-side posterior comparison), per-dimension partial missingness, per-particle contiguous gaps with SVI.

## Test plan

- [x] `scripts/test.sh` — smoke tests pass (~2 min)
- [x] `scripts/test_full.sh` — full convergence tests pass (~16 min)
- [x] Notebook runs end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dimkab:
made `unroll_missing=True` the default.

dimkab:
> b) assumes a model with independent states, but doesn't enforce this explicitly (I can run the notebook with off-diagonal terms in `A`, but it sounds like you don't want a user to do this until it is supported)

to be precise: for gaussian observations you need diagonal observation covariance (and this is enforced); the state transition covariance can be anything because the states are fully sampled anyway. The enforcement for the state noise and the initial condition covariance comes into play only for dirac observations.

dimkab:
Superseded elsewhere

PR #153, opened 2026-03-18
Add UODE Lotka-Volterra notebook

markusheinonen:
- Observe both prey and predator (H=I_2) at dt=0.2 (150 points)
- Laplace(0, 0.005) prior for SINDy-like sparsity on Theta
- SVI config: lr=1e-3, cov_rescaling=2.0, lr_decay=0.5, 2000 steps
- Correctly recovers xy interaction coefficients (Frobenius err ~0.077)
- Includes debug CLI script for systematic experiments

PR #155, opened 2026-03-19
Improved UODE inference of Lotka Volterra

mattlevine22:
Improving #153 for UODE inference of Lotka-Volterra.

Made the following changes:
- **Use Ensemble Kalman Filter** (`ContinuousTimeEnKF`) instead of Extended, as it is more numerically robust
- Removed optimization scheduler---reverted to 4000 epochs of ADAM with Ir=1e-3.
- Fixed the plotter for comparing filtered means vs data---when using `AutoDelta` guide, `svi_result.params` doesn't always do what you think it does.

Additional thoughts:
- For fully-observed, low-noise, I would recommend the `Discretizer` + `Simulator` w/ `DiracIdentityObservation` pattern, as in https://github.com/BasisResearch/dynestyx/blob/1c02f609350a98d11f3212485b8ec9601bfbe02b/docs/deep_dives/l63_speedup_dirac_vs_enkf.ipynb

N.B. I took these settings (EnKF + AutoDelta with lr=1e-3) from the SINDY-notebook for Fitz-Hugh-Nagumo. It is reassuring that they generalized nicely to this case.

N.B.B. most of the 4000 epochs are to "hyper-refine" the inference; the data are explained quite well after 1000 epochs, and the final 3000 epochs mostly focus on full sparsification (to see this, rerun with 1000 epochs).


DanWaxman (approved):
These look good to me. In the spirit of faster workflow, I'll merge these into #153

PR #163, opened 2026-03-20
Versioned docs page

mattlevine22:
- create versioned-docs (tags v0.0.1 etc, stable (most recent tag), and latest (main)).
- documentation about how versions are named / built
- utilities for building versioned docs locally

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -72,7 +72,7 @@ jobs:
           TAG: ${{ github.ref_name }}
         run: |
           VERSION="${TAG#v}"
-          uv run mike deploy --push --update-aliases --title "stable (${VERSION})" "$VERSION" stable
+          uv run mike deploy --push --update-aliases --title "$VERSION" "$VERSION" stable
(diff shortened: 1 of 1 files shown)


DanWaxman:
Okay, I think I figured out all the pain points. See https://danwaxman.github.io/dynestyx/ and let me know what you think @mattlevine22. Also, please make sure the preview script works, I have not tried it. If those look good, I'll approve and merge (though there's a few settings and manual actions I'll have to run after that).

mattlevine22:
Looks good. The preview script broke locally for me, so I removed it. Your build looks good, although I'm seeing:
- latest
- stable (0.0.3)
- stable (0.0.2)
- v0.0.1

I confirmed that v0.0.1 and latest are what they should be. 
Not so sure what "stable" is meant to be here, but I see that 0.0.3 has predict_times and 0.0.2 does not

I'm happy to accept and merge as this is better than before.

DanWaxman:
I think I fixed the stable thing. Going to merge now and change the relevant settings.

Issue #160, opened 2026-03-19
Allow Newer Jax + Python Versions


DanWaxman:
Currently, only Python <3.14 can be used, due to constraints on the JAX version. These JAX versions have themselves caused some issues, e.g., for @cooijmanstim. We should allow for more modern JAX installations, which is currently blocked by the pinned version of `cd_dynamax`. 

`uv` has an option to just ignore the pinned jax version of `cd_dynamax`, but this isn't ideal since it doesn't propagate to `pip`. A better option is to address this in `cd_dynamax` directly, which has an open PR: https://github.com/hd-UQ/cd_dynamax/pull/12.

PR #165, opened 2026-03-21
Migrate to CD-Dynamax public

mattlevine22:
- Switch to using CD-Dynamax public branch (no more `build_params` API, so integration is a bit more code)
- Added a test for SDE with non-gaussian initial condition (didn't previously support this in dynestyx, but CD-Dynamax does)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
     "equinox>=0.13.2",
     "jax>=0.6.2",
     "jaxlib>=0.6.2",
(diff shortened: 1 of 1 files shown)


mattlevine22:
> Also n.b. this is very close to resolving #160... but this tfp dependency is somewhat annoying. tfp depends on dm_tree, which is in maintenance mode and supposedly not looking to provide a 3.14 build...

Thanks! Tests now pass locally for numpyro == `0.19.0` and `0.20.0`.

PR #178, opened 2026-03-26
Add fast SDE solver backends and filter-simulator coverage.

mattlevine22:
Introduce solver modules for ODE and SDE paths, add an Euler-Maruyama scan source for SDESimulator, and update filter+simulator tests and continuous-time tutorial outputs to cover the new backend behavior.

Made-with: Cursor


DanWaxman:
With some Codex help, merged main into here. Changes seem reasonable + seems to work still.

n.b. unlike the NumPyro calls, we can actually `vmap` over these scan calls instead of doing everything in a big for loop. Worth considering.

mattlevine22:
@DanWaxman looks great thanks!
- I added a speed tip to the SDESimulator docs
- Regarding vmap vs for loop over the scans...I believe we already vmap; `SDESimulator` line 838 has `states, emissions = jax.vmap(_sim_one_trajectory)(...)`

DanWaxman:
> * Regarding vmap vs for loop over the scans...I believe we already vmap; `SDESimulator` line 838 has `states, emissions = jax.vmap(_sim_one_trajectory)(...)`

Ah, you're right, thanks! I thought we didn't do this for ODE for some reason, I guess we're just slow for discrete-time systems. :)

dynestyx/simulators.py
```diff
@@ -770,6 +720,11 @@ def __init__(
         assert self.tol_vbt < dt0, (
             "tol_vbt must be smaller than dt0 for statistically correct simulation."
         )
+        if self.source not in {"diffrax", "em_scan"}:
+            raise ValueError(
```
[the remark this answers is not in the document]
mattlevine22: Done.

dynestyx/solvers/sde.py
```diff
@@ -0,0 +1,192 @@
...
+                "solve_sde(source='em_scan') requires a numeric dt0 in "
+                "diffeqsolve_settings."
+            )
+        return _solve_sde_scan(
+            dynamics,
+            t0,
+            saveat_times,
+            x0,
+            control_path_eval,
+            float(dt0_setting),
```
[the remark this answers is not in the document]
mattlevine22: Done---introduced a `_coerce_dt` utility.

dynestyx/solvers/sde.py
```diff
@@ -0,0 +1,192 @@
...
+    x0: State,
+    control_path_eval: Callable[[Array], Array | None],
+    dt0: float,
+    *,
+    key: Array | None,
+) -> Array:
+    if key is None:
+        raise ValueError("PRNG key is required for em_scan SDE solves.")
+    if dt0 <= 0:
+        raise ValueError(f"EM scan requires dt0 > 0, got dt0={dt0}.")
```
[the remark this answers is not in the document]
mattlevine22: Done. Resulted in a larger refactor to not duplicate `early_return`. This will need further refactor to support #199, but I suggest we complete this PR as is and shift the subsequent load onto #199.

dynestyx/solvers/sde.py
```diff
@@ -100,51 +111,49 @@ def _solve_sde_diffrax(
...
     key: Array | None,
     tol_vbt: float | None,
 ) -> Array:
-    t1 = saveat_times[-1]
-
-    def _early_return():
-        return jnp.broadcast_to(x0, (len(saveat_times),) + jnp.shape(x0))
-
-    def _solve():
-        if key is None:
```
DanWaxman: Why this change?

dynestyx/solvers/sde.py
```diff
@@ -155,38 +164,51 @@ def solve_sde(
     saveat_times: Array,
     x0: State,
     control_path_eval: Callable[[Array], Array | None],
-    diffeqsolve_settings: dict[str, Any],
+    diffeqsolve_settings: dict[str, object],
```
DanWaxman: Same here

dynestyx/solvers/sde.py
```diff
@@ -155,38 +164,51 @@ def solve_sde(
...
-            tol_vbt=tol_vbt,
-        )
+    t0_arr = jnp.asarray(t0, dtype=saveat_times.dtype)
+    needs_integration = t0_arr < saveat_times[-1]
 
-    if source == "em_scan":
-        dt0_setting = diffeqsolve_settings.get("dt0")
-        if not isinstance(dt0_setting, (int, float)):
+    if key is None:
+        if bool(needs_integration):
```
DanWaxman: Is this safe? `needs_integration` is a `jax` array, right?

dynestyx/solvers/sde.py
```diff
@@ -155,38 +164,51 @@ def solve_sde(
...
-            t0,
-            saveat_times,
-            x0,
-            control_path_eval,
-            float(dt0_setting),
-            key=key,
-        )
+        return _early_return_states(x0, saveat_times)
+
+    def _do_solve(_: Array) -> Array:
```
DanWaxman, line 172: Why are we defining a function over `_`?

DanWaxman (changes requested):

dynestyx/solvers/sde.py
mattlevine22: I think this is what lax.cond needs (other option is to write some lambdas)

dynestyx/solvers/sde.py
mattlevine22: More specific, but type checkers still pass?

dynestyx/solvers/sde.py
mattlevine22: same

dynestyx/solvers/sde.py
mattlevine22: addressed this by requiring a key (this is reasonable for SDE solver, and we always feed it a key)

dynestyx/solvers/sde.py
```diff
@@ -0,0 +1,209 @@
...
+            f"{type(dynamics.state_evolution)}"
+        )
+
+    def _drift(t, y, args):
+        u_t = args(t) if args is not None else None
+        return dynamics.state_evolution.total_drift(x=y, u=u_t, t=t)
+
+    def _diffusion(t, y, args):
+        u_t = args(t) if args is not None else None
+        u_t = args(t) if args is not None else None
```
DanWaxman:
```suggestion
        u_t = args(t) if args is not None else None
```

dynestyx/utils.py
```diff
@@ -160,6 +161,31 @@ def _should_record_field(
...
+        and (
+            jnp.issubdtype(dt.dtype, jnp.integer)
+            or jnp.issubdtype(dt.dtype, jnp.floating)
+        )
+    ):
+        return float(dt)
+    raise ValueError(
+        f"{name} must be a numeric scalar (Python/NumPy real or scalar JAX array), "
+        f"got {type(dt)}."
+    )
```
DanWaxman: I kinda don't understand why we need this. If we need a float, we probably need a float everywhere and should require that from the beginning.

DanWaxman (changes requested):
Minor comments only. Other remaining thing, we should probably have a test that compares statistics of our hand-written Euler-Maruyama, on a non-trivial SDE with controls, to the results of diffrax using the Euler solver and a sufficiently small `tol_vbt` parameter.

dynestyx/utils.py
mattlevine22: I think the main use case is when it is given as a jax.Array (perhaps due to some batching needs).

The bool/Real checks are definitely a bit much, want me to remove them?

mattlevine22 (commented):
Fixed the 1 minor correction and responded to use case of `_coerce_dt`

dynestyx/utils.py
DanWaxman: As discussed, I guess my main objection is that we are concretizing a jax array in the first place. I think this will break things if traced. If we can at all (i) store as a float from the beginning, and only convert to an array as-needed, or (ii) always store as a jax array and deal with the consequences of that, I think the code will be more robust.

mattlevine22: I looked more closely, and I think the main reason we have floats *anywhere* is for friendly user API (no one wants to write `dt0=jnp.array(0.01)` ).

What do you think about doing float -> Array at a high-level early on and then do our internals all on Arrays?

DanWaxman:
I'd be happy with conversion at init :).

DanWaxman (approved):

PR #183, opened 2026-03-31
test and fix log softmax stability in particles_to_delta_mixtures

cooijmanstim:
This fixes an issue with the log softmax computation for large negative and uniform log probabilities. In isolation the issue is that these two expressions return very different values:

```
w = -1e10 * jnp.ones([11])
w_max = w.max()
print((w - w_max) - jnp.log(jnp.sum(jnp.exp(w - w_max))))
print(w - (w_max + jnp.log(jnp.sum(jnp.exp(w - w_max)))))
```

The latter returns all zeros, which is not a normalized log probability vector.


DanWaxman (approved):
This looks great, thanks Tim!

PR #188, opened 2026-04-06
add version property and bump version to ship to pypi

mattlevine22:
Ensure `import dynestyx; print(dynestyx.__version__)` works.


DanWaxman (approved):
Cool! I'm fine with merging this, though perhaps we'd like to wait for #173 to land before bumping a new version? I'll be finished with the documentation tasks there today.

PR #189, opened 2026-04-06
fix formatting of filter config docstrings

mattlevine22:
Avoids syntax warnings during import


DanWaxman (approved):
I made sure these render well (+ added HMM to API docs, which was missing previously). Looks good to me, thanks Matt!

PR #194, opened 2026-04-09
Better Document Filter Sources

### LINKED ISSUE #122: Better Document Filter Sources
Right now, the user has to do a bit of guessing to figure out the filter source. This should be documented better in the configs.

DanWaxman:
This commit/PR better documents filter sources by using different literal classes in the typing of each dataclass. For example, `ContinuousTimeDPFConfig` now has field `filter_source: CDDynamaxOnlyFilterSource`.

Resolves #122


DanWaxman:
Yeah... I don't have a much better way of documenting this well, unfortunately. Other than maybe having some dummy mixin class of `CuthbertFilter`/`DynestyxFilter`/`CDDynamaxFilter`, which I think is also better than we have, but worse than this.

mattlevine22 (approved):
satisfaction => action!

PR #197, opened 2026-04-13
Fix FAQ for hierarchical models

DanWaxman:
Updating FAQs from #173. Checked the links worked.


mattlevine22 (approved):
links work looks good, thanks!

Issue #115, opened 2026-02-25
Ensure Euler-Maruyama compatibility with Gaussian-transition filters


mattlevine22:
1. Check compatibilities
2. Does the EM-discrete-evolution need to be a special class to be recognized for structured inference? If so, fix this.

DanWaxman:
One issue with this: `GaussianStateEvolution` requires a static covariance, but Euler-Maruyama plainly gives time-varying covariance (for non-equally spaced observation times). @mattlevine22 do you remember if this was for dynamax reasons?

PR #199, opened 2026-04-14
Make Discretizer use GaussianStateEvolution for increased filter compatibility

mattlevine22:
Addresses issue #115 

- Gives `GaussianStateEvolution` a time-varying covariance field (or jax.array)
- CD_Dynamax/Dynamax integration with `GaussianStateEvolution` **errors** if covariance is a callable
- Refactor `discretizers.py` to build a `class EulerMaruyamaGaussianStateEvolution(GaussianStateEvolution)`
- Added `tests_discretizers.py` which runs Cuthbert EKF with a discretized SDE model.

N.B. CD_Dynamax/Dynamax discrete-time integrations don't support time-varying `F(, .. t)` anyway

N.B.This PR only adds support for time-varying `GaussianStateEvolution` to enable discretized + discrete-time-filter patterns. It does NOT modify: [`LinearGaussianStateEvolution`, `GaussianObservation`, `LinearGaussianObservation`],


dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -408,6 +408,13 @@ def gaussian_to_nlgssm_params(dynamics: DynamicalModel) -> ParamsNLGSSM:
     state_dim = dynamics.state_dim
     control_dim = dynamics.control_dim
 
+    if callable(evo.cov):
+        raise TypeError(
+            "cd_dynamax discrete EKF/UKF requires array-valued process covariance. "
+            "Received callable state_evolution.cov. Use a fixed covariance matrix, "
+            "or choose a backend that supports callable covariances."
+        )
+
```
DanWaxman, line 417: Do we also have to do this for LGSSM?

dynestyx/discretizers.py
```diff
@@ -32,91 +32,134 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
+    cte: ContinuousTimeStateEvolution,
+    x,
+    u,
+    t_now,
+    t_next,
+):
+    """Euler–Maruyama: mean `x + drift*dt`, cov `(L@Q@L.T)*dt` with `Q=I`.
+
+    Supports batched time (and optional control) matching the previous
+    `EulerMaruyamaGaussianStateEvolution` behavior.
```
DanWaxman: de-AI-ify pls :)

dynestyx/discretizers.py
```diff
@@ -32,91 +32,134 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
+
+    Returns:
+        dict: {
+            "loc": ndarray
+                Mean of next state(s): shape (dim_state,) or (num_timepoints, dim_state)
+            "cov": ndarray
+                Covariance of next state(s): shape (dim_state, dim_state) or (num_timepoints, dim_state, dim_state)
```
DanWaxman: if this allows batched shapes, this should be reflected in the docstring array shapes

dynestyx/discretizers.py
```diff
@@ -32,91 +32,134 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
+            )["cov"],
+        )
 
     def __call__(self, x, u, t_now, t_next):
+        """Single-pass transition; avoids calling ``F`` and ``cov`` separately.
+
+        The parent `GaussianStateEvolution.__call__` would evaluate the
+        Euler–Maruyama drift and diffusion twice. Under `jax.vmap` (e.g.
+        plate-batched cuthbert EKF), that split can change tracing/shapes. This
+        override matches the original one-step implementation.
```
DanWaxman: same here

dynestyx/discretizers.py
```diff
@@ -32,91 +32,134 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
-
-            Returns:
-                dist: MultivariateNormal distribution
-                    - loc: (dim_state, num_timepoints) or (dim_state)
-                    - covariance_matrix: (dim_state, dim_state, num_timepoints) or (dim_state, dim_state)
-        """
+        em_result = _euler_maruyama_loc_cov(self.cte, x, u, t_now, t_next)
+        return dist.MultivariateNormal(
+            loc=em_result["loc"], covariance_matrix=em_result["cov"]
+        )
```
DanWaxman: I'm not so clear on the benefit of making this a helper function, to be honest. It seems like it can be reasonably in-lined here.

dynestyx/discretizers.py
```diff
@@ -32,91 +32,134 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
+    return {"loc": loc, "cov": cov}
+
+
+class EulerMaruyamaGaussianStateEvolution(GaussianStateEvolution):
+    """Euler–Maruyama discretization as `GaussianStateEvolution`.
+
+    Holds ``cte`` as an explicit field so `DynamicalModel` pytrees under
+    `numpyro.plate` still expose batched continuous-time parameters for
+    simulator slicing (closures over ``cte`` alone would hide those arrays).
+
```
DanWaxman: This seems a bit in-the-weeds. Would prefer to just write that we hold `cte` for pytree compatibility.

DanWaxman (changes requested):
I think this looks good in principle, maybe with some style comments. This should've run the hierarchical discretized smoke tests, but it would also be nice to run the `08_hierarchical_inference.ipynb` tutorial notebook and ensure there's no qualitative change there (I don't expect there is).

dynestyx/solvers/sde.py
```diff
@@ -27,10 +35,293 @@ def _apply_diffusion(diffusion_term: Array, dw: Array) -> Array:
...
+def _bm_dim_or_default(state_evolution: ContinuousTimeStateEvolution) -> int:
+    """Return Brownian dimension, defaulting to 1 when unspecified.
+
+    Args:
+        state_evolution: Continuous-time state evolution.
+
+    Returns:
+        Brownian motion dimension used by EM sampling.
+    """
+    return int(state_evolution.bm_dim) if state_evolution.bm_dim is not None else 1
```
DanWaxman: Why don't we just set the corresponding `state_evolution.bm_dim`? Seems like needless indirection.

dynestyx/solvers/sde.py
```diff
@@ -27,10 +35,293 @@ def _apply_diffusion(diffusion_term: Array, dw: Array) -> Array:
...
+
+
+def euler_maruyama_step_loc_cov(
+    state_evolution: ContinuousTimeStateEvolution,
+    x: Array,
+    u: Array | None,
+    t_now: Array,
+    dt: Array,
+) -> tuple[Array, Array]:
+    """Compute one EM moment step over a fixed `dt`.
```
DanWaxman: This feels like a particularly simple function.

dynestyx/inference/integrations/cd_dynamax/utils.py
mattlevine22: No, because currently `LinearGaussianStateEvolution` does not support time-varying parameters (A/B/b/cov).

dynestyx/discretizers.py
mattlevine22: done

dynestyx/discretizers.py
mattlevine22: done

dynestyx/discretizers.py
mattlevine22: done

dynestyx/discretizers.py
mattlevine22: done

dynestyx/discretizers.py
mattlevine22: We want it to construct F  and cov in the `__init__` method, as well as use it in the `__call__` method

mattlevine22:
This PR only adds support for time-varying `GaussianStateEvolution` to enable discretized + discrete-time-filter patterns. It does not modify things like: [`LinearGaussianStateEvolution`, `GaussianObservation`, `LinearGaussianObservation`],

dynestyx/inference/integrations/cd_dynamax/utils.py
DanWaxman: I see, this is the job of #200.

dynestyx/solvers/sde.py
```diff
@@ -27,10 +35,293 @@ def _apply_diffusion(diffusion_term: Array, dw: Array) -> Array:
...
+          input or `(num_timepoints, state_dim)` for batched-time input.
+        - `"cov"`: predicted covariance matrix/matrices, shape
+          `(state_dim, state_dim)` for single-transition input or
+          `(num_timepoints, state_dim, state_dim)` for batched-time input.
+
+    Batching behavior:
+        Batched mode maps across the time axis, pairing
+        `x[:, i], u[:, i], t_now[i], t_next[i]` for each `i`.
+        Scalar inputs are promoted to a batch of size 1 internally and squeezed
+        back to single-transition outputs.
```
DanWaxman: Isn't this counter to other places?

dynestyx/utils.py
```diff
@@ -160,6 +161,31 @@ def _should_record_field(
...
+            or jnp.issubdtype(dt.dtype, jnp.floating)
+        )
+    ):
+        return float(dt)
+    raise ValueError(
+        f"{name} must be a numeric scalar (Python/NumPy real or scalar JAX array), "
+        f"got {type(dt)}."
+    )
+
+
```
DanWaxman: I think we got rid of this in #178 ?

dynestyx/discretizers.py
```diff
@@ -32,91 +32,66 @@ def _ensure_ctse_bm_dim(dynamics: DynamicalModel) -> DynamicalModel:
...
 
-class _EulerMaruyamaDiscreteEvolution(DiscreteTimeStateEvolution):
-    """x_{t+1} ~ N(x + drift*dt, (L@Q@L.T)*dt)."""
+class EulerMaruyamaGaussianStateEvolution(GaussianStateEvolution):
+    """Euler–Maruyama discretization as `GaussianStateEvolution`.
+
+    Holds ``cte`` as an explicit field for Equinox pytree compatibility.
+
+    ``F`` and ``cov`` are optional constructor args so Equinox/dataclass-style
+    but we don't use them.
```
DanWaxman: ?

dynestyx/solvers/sde.py
DanWaxman: Also I don't really see how 1 is a reasonable default. If anything, we should raise or set up a probe state pattern to get the right shape.

DanWaxman (changes requested):

dynestyx/utils.py
mattlevine22: good catch, removed, thanks

dynestyx/solvers/sde.py
mattlevine22: good catch, agreed. Changed it to just error if state_evolution.bm_dim is None (should already be set upstream).

dynestyx/solvers/sde.py
mattlevine22: ok agreed, I removed `euler_maruyama_step_loc_cov` and `euler_maruyama_step_sample`

dynestyx/solvers/sde.py
mattlevine22: Yes, and turned out to be a bit of a rabbit hole that I think is reasonably addressed by commit 9a14046.


Codex's cheeky discription of the commit:

Made the Euler-Maruyama discretization path follow the project’s standard batching contract: euler_maruyama_loc_cov(...) now treats batched states as (..., state_dim) and preserves leading batch/plate dimensions instead of relying on the old batch-last (state_dim, T) convention. To keep the rest of the stack consistent, I updated the DiscreteTimeSimulator Dirac-conditioning path to pass time-batched states in that same layout, fixed DiracIdentityObservation so it uses event_dim=0 for scalar states and event_dim=1 for vector states, and added regression coverage for direct EM batching, discretized Dirac observations, and hierarchical single-/nested-plate Dirac behavior with and without Discretizer.

mattlevine22 (commented):
agreed with your feedback---some were easy to address, but the tricky one was making `euler_maruyama_loc_cov` adhere to (..., T, state_dim) while still working for the Perfect-Observation case. Had to do some finagling to get it to work, but I think things are consistent.

One thing that came up is that we allow the state x to be () or (state_dim,)....would be great if we had a good way of normalizing this aspect. I tried, but got messy on first pass so I left it alone.

docs/tutorials/gentle_intro/03_filtering_mll.ipynb
```diff
@@ -183,9 +183,10 @@
...
-        "\n",
-        "We'll now build a function that returns the MLL at a particular value of the parameter $\\rho$, conditioned on data, and plot the resulting *likelihood profile*."
+        "# To condition a model using a filtering algorithm, we use a `Filter` object, which takes a `FilterConfig` from `dynestyx.inference.filters` that specifies which filtering algorithm to use.  \n",
+        "# In this notebook, we'll compare several available options: the ensemble Kalman filter (EnKF, well-suited for non-linear Gaussian models and used by default), the particle filter (PF, more computationally expensive but fully general), and the extended Kalman filter (EKF, less accurate for strongly non-linear systems, but efficient and optimal for linear-Gaussian models).\n",
+        "#\n",
+        "# Next, we'll build a function that evaluates and returns the MLL at a fixed value of the parameter $\\rho$, conditioned on data, and use it to plot a *likelihood profile*."
```
DanWaxman: ?

dynestyx/solvers/sde.py
```diff
@@ -27,10 +35,270 @@ def _apply_diffusion(diffusion_term: Array, dw: Array) -> Array:
...
+
+    loc_state_dim = loc_flat.shape[-1]
+    if loc_state_dim != state_dim:
+        raise ValueError(
+            "euler_maruyama_loc_cov produced a location shape inconsistent with "
+            f"the input state shape: x.shape={x_arr.shape}, loc.shape={loc_flat.shape}."
+        )
+    loc = loc_flat.reshape(batch_shape + (loc_state_dim,))
+    cov = cov_flat.reshape(batch_shape + cov_flat.shape[-2:])
+    return {"loc": loc, "cov": cov}
```
DanWaxman, line 298: Would it not be a bit clearer to just vmap multiple times? This seems correct to me, but flattening and reshaping seems like lots of memory shenanigans and difficult to read.

dynestyx/discretizers.py
DanWaxman: Bump

dynestyx/simulators.py
```diff
@@ -970,25 +970,16 @@ def _simulate(
...
-            u_prev_batch_last = (
-                jnp.swapaxes(u_prev, 0, 1) if u_prev is not None else None
-            )
-
             with numpyro.plate("time", T - 1):
                 trans = dynamics.state_evolution(
-                    x_prev_batch_last,
-                    u_prev_batch_last,
+                    x_prev,
+                    u_prev,
```
DanWaxman: This should probably be vmapped

DanWaxman (changes requested):

dynestyx/discretizers.py
mattlevine22: oops, removed that and left a 1-line docstring

docs/tutorials/gentle_intro/03_filtering_mll.ipynb
mattlevine22: thanks for catching, uncommented

dynestyx/solvers/sde.py
mattlevine22: sure, switched to that. shorter and easier to read!

dynestyx/simulators.py
mattlevine22: done

DanWaxman (approved):
Looks great, thanks!!

mattlevine22:
Closes #200

PR #192, opened 2026-04-09
Add Smoothers

DanWaxman:
This PR adds smoothers under a new `Smoother(...)` handler, mirroring the current `Filter(...)` interface. It introduces a number of different smoothing algorithms from CD-Dynamax/cuthbert, some new documentation, and also introduces a bit of a refactor to integration code that was getting quite long (now split into filtering, smoothing, and util files).

Currently, prediction is not handled the way I would like it to be: I would like prediction to use the smoother within the smoothing window, and only pass forward the final predict_times as a filtered_times/filtered_dists pair. This is not currently done because missing data is not properly implemented in cd_dynamax for continuous time, and our indexing is not really correct for discrete time. For now, we decided to properly handle `predict_times >= max(obs_times)`, and defer a more complete treatment to a later time.

- [x] Include Smoother Handler
- [x] Integration
  - [x] Cuthbert
    - [x] Particle Smoother
    - [x] Kalman (RTS) Smoother
    - [x] Extended Kalman (RTS) Smoother
  - [x] CD-Dynamax
    - [x] Kalman (RTS) Smoother
    - [x] Extended Kalman (RTS) Smoother
  - [x] Dynamax
    - [x] Kalman (RTS) Smoother
    - [x] Extended Kalman (RTS) Smoother
    - [x] Unscented Kalman Smoother 
- [x] New tests
    - [x] Smoke tests
    - [x] Science tests
- [x] Documentation
    - [x] API Reference Pages
    - [x] Full docstring & references for configs
    - [x] Tutorial notebooks
- [x] Proper prediction behavior

PR #209, opened 2026-04-22
Add Cuthbert EnKF

DanWaxman:
There is now an EnKF in cuthbert, as of v0.0.10. This commit adds integrations to this filter, including:
- Upgrading the required cuthbert version
- Updating the corresponding filter_config information
- Make EnKF the default discrete-time filter
- Add EnKF integration code
- Update tests accordingly
- Add to documentation (including discrete-time LTI profile likelihood notebook)

Remaining: 
- [x] Let's run this on some real examples and really check that this is better than EKF in order to be the default.


docs/faq.md
```diff
@@ -97,7 +97,7 @@ Hierarchical models are supported by the [`dsx.plate`](./api_reference/public/ha
 
 ## What about neural nets?
 
-We will put examples up soon. See [CD-Dynamax's Lorenz 63 neural drift tutorial](https://github.com/hd-UQ/cd_dynamax/blob/dev-numpyro-api/demos/numpyro/notebooks/lorenz63_nndrift_sgd_fit_to_data_tutorial_newAPI.ipynb) to convince yourself that this will work well.
+See our [neural drift correction deep dive](deep_dives/neural_drift_correction.ipynb). It shows how to keep the observation model and filter workflow fixed while swapping a mechanistic ODE drift for a mechanistic-plus-neural residual drift.
```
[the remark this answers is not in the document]
DanWaxman: @copilot Revert this change entirely, actually. It was made by accident.

dynestyx/inference/integrations/cuthbert/discrete.py
```diff
@@ -231,6 +255,111 @@ def log_potential(x_prev, x, mi: CuthbertInputs):
...
+                    or (
+                        isinstance(edist, dist.Independent)
+                        and isinstance(edist.base_dist, dist.Normal)
+                    )
+                ):
+                    raise TypeError(
+                        "cuthbert EnKF observation callable must keep returning "
+                        "Gaussian distributions; got "
+                        f"{type(edist).__name__}."
+                    )
```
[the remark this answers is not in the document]
DanWaxman: Added some checks that the chol_R is not state-dependent. Time-varying is handled just fine, I think.

docs/api_reference/public/inference/filter_configs.md
```diff
@@ -7,9 +7,9 @@ The single `Filter()` handler is directed to the appropriate filtering algorithm
 | Config class               | Time domain         | When it fits best |
 |----------------------------|---------------------|-------------------|
 | `KFConfig`                 | Discrete            | Linear-Gaussian dynamics and linear-Gaussian observations (exact & optimal). |
-| `EKFConfig`                | Discrete            | Nonlinear, differentiable Gaussian dynamics, nonlinear but differentiable Gaussian observations (approximate). *(default)*. |
+| `EnKFConfig`               | Discrete            | Nonlinear or expensive models with Gaussian observations; cuthbert-backed and a good general-purpose default. *(default)* |
+| `EKFConfig`                | Discrete            | Nonlinear, differentiable Gaussian dynamics, nonlinear but differentiable Gaussian observations (approximate). |
```
mattlevine22: Can actually run EKF with non-gaussian observations via Cuthbert

docs/deep_dives/discrete_time_lti_profile_likelihood.ipynb
mattlevine22, line 1: Image

Adjust ylimit axis for profile likelihood timers.

DanWaxman: Are you able to re-run this? I think my laptop is just eepy today...

dynestyx/inference/integrations/cuthbert/discrete.py
```diff
@@ -158,17 +229,18 @@ def run_discrete_filter(
 
     if isinstance(filter_config, PFConfig):
         _add_sites_pf(name, states, record_kwargs)
-        particles = states.particles
+        particles = states.particles[1:]
```
mattlevine22: is this `[1:]` (and subsequents) catching a previous off-by-one error??

DanWaxman:
Oops yes, sorry meant to write that in the description. Notably, things are
already sliced when adding sites.

dynestyx/inference/integrations/cuthbert/discrete.py
mattlevine22: If so, it seems like this problem should be managed upstream, e.g. in `compute_cuthbert_filter` or even `cuthbert_filter`.

Currently, the batched filter computations would not pass through the above changes (but does use `compute_cuthbert_filter`).

dynestyx/inference/integrations/cuthbert/discrete.py
```diff
@@ -231,6 +303,107 @@ def log_potential(x_prev, x, mi: CuthbertInputs):
...
+        probe_u = jnp.zeros(())
+        probe_t = jnp.zeros(())
+        try:
+            probe_d0 = obs_model(jnp.zeros((state_dim,)), probe_u, probe_t)
+            probe_d1 = obs_model(jnp.ones((state_dim,)), probe_u, probe_t)
+        except Exception:
+            probe_d0 = probe_d1 = None
+        if probe_d0 is not None and probe_d1 is not None:
+            chol0 = _extract_gaussian_chol(probe_d0, obs_dim)
+            _check_state_independent_noise(chol0, probe_d1, obs_dim)
```
mattlevine22: would be more readable as a utility function.

dynestyx/inference/integrations/cuthbert/discrete.py
```diff
@@ -472,8 +647,7 @@ def _add_sites_taylor_kf(
         numpyro.deterministic(f"{name}_filtered_states_chol_cov", chol_cov)
 
     if add_filtered_states_cov or add_filtered_states_cov_diag:
-        chol_t = jnp.transpose(chol_cov, (0, 2, 1))
-        filtered_cov = jnp.matmul(chol_cov, chol_t)
+        filtered_cov = covariance_from_cholesky(chol_cov)
 
     if add_filtered_states_cov:
         numpyro.deterministic(f"{name}_filtered_states_cov", filtered_cov)
```
mattlevine22, line 684: At 650 lines and supporting 4 different filters and their own utilities (and some shared), I think this file deserves some refactoring into multiple files.

dynestyx/inference/integrations/cuthbert/discrete.py
mattlevine22: And is there a test that would have failed before but now can pass? Wondering how we missed this...

mattlevine22 (commented):
Main thoughts are:
1. There were some off-by-one fixes `[1:]`...
- how did we miss these? Can we have a test to protect us?
- To what extent can this be handled higher up in the Cuthbert integration  (e.g. `cuthbert_filter`)?
- And is this fixed for the batched case as well (different branches of filters it seems)

2. The `cuthbert/discrete.py` file has gotten really long and tough to follow.
- Would highly recommend refactoring into multiple files and some shared utilities.

dynestyx/inference/integrations/cuthbert/discrete.py
DanWaxman: > And is there a test that would have failed before but now can pass? Wondering how we missed this...

Yup, added a test for this `test_filters.py::test_cuthbert_filtered_distribution_shapes_match_observations`. The error was mostly innocuous, which is why we didn't see it before. This only gets passed forward as `filtered_dists`, and it didn't affect the prediction behavior at all because of how scans work.

> If so, it seems like this problem should be managed upstream, e.g. in compute_cuthbert_filter or even cuthbert_filter.

Yeah, that seems fair.

dynestyx/inference/integrations/cuthbert/discrete.py
DanWaxman: Agreed. I started this actually in #192. I think it makes sense to continue the refactoring there.

dynestyx/inference/integrations/cuthbert/discrete.py
DanWaxman: Done!

dynestyx/inference/integrations/cuthbert/discrete.py
DanWaxman: Done

DanWaxman:
> Main thoughts are:
> 1. There were some off-by-one fixes `[1:]`...
>   * how did we miss these? Can we have a test to protect us?

The issue was mostly innocuous, it just got passed forward to `predict_times` where it didn't really do anything. We were always using the logged states, I think, which were properly indexed.

>   * To what extent can this be handled higher up in the Cuthbert integration  (e.g. `cuthbert_filter`)?

This is a good point, handled in [7e3c3fc](https://github.com/BasisResearch/dynestyx/pull/209/commits/7e3c3fcf7016bb277a06a9d93e5c3800308b7764). We don't have access to `cuthbert_filter` directly (that's a function in `cuthbert`), but it's now handled in `compute_cuthbert_filter`. 

>   * And is this fixed for the batched case as well (different branches of filters it seems)

Yup, also done by doing this in `compute_cuthbert_filter`

> 2. The `cuthbert/discrete.py` file has gotten really long and tough to follow.
>   * Would highly recommend refactoring into multiple files and some shared utilities

Agreed. This is part underway in #192, which I think is a more natural place to undertake this refactoring.

mattlevine22 (approved):
Looks good. 
- Confirmed that Tutorial docs (NUTS + SVI) produce similar outputs (and speed) with EnKF and EKF
- Confirmed that test_science for discretized L63 produces similar posteriors w/ NUTS + EKF and NUTS + EnKF

PR #211, opened 2026-04-24
Add Associative KF

DanWaxman:
Cuthbert supports associative Kalman filtering, which should just work as-advertised. This PR adds it as an option, some test coverage, and a small benchmarking notebook.


dynestyx/inference/integrations/cuthbert/discrete.py
```diff
@@ -229,7 +229,19 @@ def compute_cuthbert_filter(
             "Expected KFConfig, EKFConfig, EnKFConfig, PFConfig."
         )
 
-    raw_states = cuthbert_filter(filter_obj, cuthbert_inputs, parallel=False, key=key)
+    parallel = isinstance(filter_config, KFConfig) and filter_config.associative
+    if parallel and not filter_obj.associative:
```
mattlevine22, line 233: I guess this error check is for internal implementation stuff? Hard to see how you can hit this error since filter_config -> filter_kwargs -> filter_obj and parallel is IFF KFConfg

dynestyx/inference/filter_configs.py
```diff
@@ -333,6 +333,9 @@ class KFConfig(BaseFilterConfig):
 
     Attributes:
         filter_source (FilterSource): Backend. Defaults to `"cd_dynamax"`.
+        associative (bool): Whether to enable cuthbert's associative
+            parallel-in-time scan. This is only supported when
+            `filter_source="cuthbert"`. Defaults to `False`.
```
mattlevine22: Defaults to True

dynestyx/inference/integrations/cuthbert/discrete.py
mattlevine22: Oh does this come up specifically bc of CD-Dynamax stuff?

mattlevine22 (approved):
Approved w small requested changes:

1. Docstring says Defaults to True, gotta swap that that.
2. I commented about the error message, but I think it's fine. Could be confusing to a user why they got an error if they used CD-Dynamax KF though (it doesn't explain to them that's the reason).

dynestyx/inference/integrations/cuthbert/discrete.py
DanWaxman: In principle, we could add associative scans for some other filters in the future. Support for a particular filter being associative is via its `filter_obj.associative` property is true -- it is an internal check, but I think a nice one to have.

dynestyx/inference/filter_configs.py
DanWaxman: Updated, thanks

Issue #212, opened 2026-04-24
Allow diffusion to be specified as a constant


mattlevine22:
Currently `ContinuousTimeStateEvolution` requires `diffusion` to be `lambda x, u, t: ` Callable.

1. This makes writing models with constant diffusion (kinda the default use case) unnecessarily cumbersome
2. Makes it harder for us to exploit things like scalar / diagonal diffusion coefficients.

Could do some nice auto-broadcasting for hierarchical cases as well here.

DanWaxman:
Resolved by #214

Issue #16, opened 2026-01-09
User-friendly classes for `state_evolution` and `observation_model`


mattlevine22:
Think about user experience for building `dynestyx.DynamicalModel` objects. Namely, consider making factories that render simple specifications to `equinox` modules.

mattlevine22:
Let's start checking signatures of observation_model and state_evolution (and other DynamicalModel stuff).

PR #214, opened 2026-04-27
Support structured diffusion specs in continuous-time models and enforce cd-dynamax full-diffusion constraints

mattlevine22:
Addresses #212

## Summary

This PR introduces a structured diffusion API for continuous-time models.

The main change is a new `dynestyx.models.diffusions` module with first-class diffusion objects:
- `Diffusion`
- `FullDiffusion`
- `DiagonalDiffusion`
- `ScalarDiffusion`

These objects make the shape and interpretation of the SDE diffusion term explicit, instead of relying on loosely structured `diffusion_coefficient` inputs.

In particular, this PR:
- adds structured diffusion classes for full, diagonal, and scalar diffusion coefficients
- updates `ContinuousTimeStateEvolution` / `DynamicalModel` to canonicalize and validate diffusion metadata centrally
- clarifies `bm_dim` behavior:
  - `FullDiffusion`: `bm_dim` is inferred from trailing matrix shape when omitted
  - `DiagonalDiffusion` / `ScalarDiffusion`: `bm_dim` is explicit and must be `1` or `state_dim`
- updates SDE solver internals so Euler–Maruyama moments, sampling, and the Diffrax solve path all use the same diffusion semantics
- updates continuous cd-dynamax integration to validate diffusion structure up front and fail early on unsupported cases
- updates `LTI_continuous` and related examples/docs to use the new diffusion API directly
- expands coverage across model validation, discretizers, `bm_dim`/plate behavior, and hierarchical smoke tests

## Why

This change makes continuous-time model specification clearer and safer.

It also improves performance in simulation-oriented SDE paths when the diffusion has scalar or diagonal structure, since we can avoid treating every diffusion as a dense full matrix. The speedup figure below illustrates this benefit.

Just as importantly, invalid setups now fail earlier with clearer errors, especially in continuous cd-dynamax filtering where only square full diffusion is currently supported.

## Notes

- Continuous cd-dynamax filters currently require square full diffusion (`bm_dim == state_dim`).
- More general rectangular diffusion remains supported in simulation-oriented SDE / Euler–Maruyama paths, but not in the current continuous cd-dynamax filter backend.
- I explored exploiting scalar/diagonal structure more aggressively inside `Filter + Discretizer`, but did not see meaningful gains there, so those backend-specific optimizations were not kept; will matter more with multi-step discretizers, and can leverage then.

## Tests

Added or updated coverage in:
- `tests/test_models_core.py`
- `tests/test_discretizers.py`
- `tests/test_bm_dim_plate.py`
- `tests/test_hierarchical_simulator_discretizer_smokes.py`

Below image showing speed-ups of solver using scalar/diag/full diffusion coefficient in SDE.
image


dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -22,40 +22,73 @@
...
+        u=None,
+        t=jnp.array(0.0),
+        state_dim=state_dim,
+    )
+    bm_dim = diffusion_as_matrix(diffusion, state_dim=state_dim).shape[-1]
+    if bm_dim > state_dim:
+        raise ValueError(
+            "Continuous cd-dynamax filters require bm_dim <= state_dim. "
+            f"Got state_dim={state_dim}, bm_dim={bm_dim}."
+        )
```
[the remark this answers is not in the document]
mattlevine22: @copilot update the docs/descriptions to match the current behavior.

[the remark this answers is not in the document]
mattlevine22: sure

DanWaxman:
Thanks, this looks interesting!!

> Note that I tried to exploit in Filter + Discretizer, but didn't see any speedups so I removed those implementations.

Out of curiosity, did you try for state estimation only or for parameter inference?

mattlevine22:
> Thanks, this looks interesting!!
> 
> > Note that I tried to exploit in Filter + Discretizer, but didn't see any speedups so I removed those implementations.
> 
> Out of curiosity, did you try for state estimation only or for parameter inference?

Only did a forward pass of the filter; didn't use in parameter inference, and actually didn't even look at expense of gradients. Would you expect those to be noticeably different?

mattlevine22:
> > Thanks, this looks interesting!!
> > > Note that I tried to exploit in Filter + Discretizer, but didn't see any speedups so I removed those implementations.
> > 
> > 
> > Out of curiosity, did you try for state estimation only or for parameter inference?
> 
> Only did a forward pass of the filter; didn't use in parameter inference, and actually didn't even look at expense of gradients. Would you expect those to be noticeably different?



> > Thanks, this looks interesting!!
> > > Note that I tried to exploit in Filter + Discretizer, but didn't see any speedups so I removed those implementations.
> > 
> > 
> > Out of curiosity, did you try for state estimation only or for parameter inference?
> 
> Only did a forward pass of the filter; didn't use in parameter inference, and actually didn't even look at expense of gradients. Would you expect those to be noticeably different?

@DanWaxman I looked at value_and_grad, but saw inconsistent improvements. The best I saw was like a 5% gain. I think this is because currently, we only do this multiply once-per-observation (and we have other operations that occur per-observation that dominate). When we have fancier discretization that take many steps between observations, it will be more impactful I think.


I'd recommend for now introducing the API and leveraging in the simple SDE solver where we know it clearly helps (esp. in large dims), then leveraging diagonal structure as it becomes usable on the methods end.

DanWaxman:
> @DanWaxman I looked at value_and_grad, but saw inconsistent improvements. The best I saw was like a 5% gain. I think this is because currently, we only do this multiply once-per-observation (and we have other operations that occur per-observation that dominate). When we have fancier discretization that take many steps between observations, it will be more impactful I think.
> 
> I'd recommend for now introducing the API and leveraging in the simple SDE solver where we know it clearly helps (esp. in large dims), then leveraging diagonal structure as it becomes usable on the methods end.

That makes sense!

dynestyx/models/diffusions.py
```diff
@@ -0,0 +1,206 @@
...
+    return jnp.asarray(value)
+
+
+def resolve_diffusion_metadata(
+    shape: tuple[int, ...],
+    *,
+    state_dim: int,
+    diffusion_type: DiffusionType | None,
+    bm_dim: int | None,
+) -> tuple[DiffusionType, int]:
```
DanWaxman: I think it makes more sense to resolve this diffusion type at model initialization, rather than at evaluation time. This comes with the disadvantage of not allowing diffusion type to change spatiotemporally, but that is something I'm happy to live with, personally.

DanWaxman (changes requested):

dynestyx/models/diffusions.py
DanWaxman: (In general, I'd prefer our inference code to be simple, and initialization code be complex. This mirrors the "dream scenario" where we're doing program introspection to mark things, for instance.)

mattlevine22: great catch thanks. Now it is only used in `__init__`.

DanWaxman (commented):
I find all this `evaluate_diffusion(...)` business pretty awkward, to be honest. Namely, I find it awkward that we have to manually go "into" `state_evo` and pick out `state_evo.diffusion_coefficient`, `state_evo.diffusion_type`, and `state_evo.bm_dim` every time -- this metadata only makes sense as a package, so we shouldn't have to pick out all of them every time.

Do you have thoughts on putting these into a single dataclass (i.e., have a `Diffusion(eqx.Module)`) that packages all this info together? Then, we can just have something like`diffusion.evaluate_diffusion(x=...,u=...,t=...)`.

mattlevine22:
I like this idea and gets at #16.

Could also have some constructor classes:

```python

ContinuousTimeStateEvolution(
    drift=...,
    diffusion=Diffusion.diag(jnp.array([0.1, 0.2]), bm_dim=2),
)

class Diffusion(eqx.Module):
    coefficient: DiffusionSpec
    diffusion_type: DiffusionType | None = eqx.field(static=True, default=None)
    bm_dim: int | None = eqx.field(static=True, default=None)

    @classmethod
    def full(cls, coefficient, bm_dim: int | None = None):
        return cls(coefficient=coefficient, diffusion_type="full", bm_dim=bm_dim)

    @classmethod
    def diag(cls, coefficient, bm_dim: int):
        return cls(coefficient=coefficient, diffusion_type="diag", bm_dim=bm_dim)

    @classmethod
    def scalar(cls, coefficient, bm_dim: int):
        return cls(coefficient=coefficient, diffusion_type="scalar", bm_dim=bm_dim)
```

mattlevine22:
Could also do this for Drifts...and I think this may help us a lot down the road in adding new features.

May want to think ahead a bit more around names to anticipate future dev

docs/api_reference/public/models/core/diffusion.md
```diff
@@ -0,0 +1,233 @@
...
+Dynestyx exposes three structured diffusion classes:
+
+- `FullDiffusion`: general matrix-valued diffusion \(L \in \mathbb{R}^{d_x \times d_w}\)
+- `DiagonalDiffusion`: vector-valued diffusion \(v \in \mathbb{R}^{d_x}\)
+- `ScalarDiffusion`: scalar-valued diffusion \(\sigma \in \mathbb{R}\)
+
+In practice, many models use a constant diffusion coefficient. In those cases,
+pass the matrix/vector/scalar value directly. Reserve callable diffusion
+coefficients for cases where the coefficient genuinely depends on state,
+control, or time.
```
DanWaxman, line 33: Examples might be helpful here

DanWaxman: Or, write that you're about to give examples :). Though these may be more useful in th eopposite direction than as-written now (i.e., "if your mathematical form is X, use code object Y" instead of "use code object Y if you have mathematical form X").

docs/logo/make_logo.py
DanWaxman, line 1: Actually not sure if this file still runs lol

docs/math_intro.md
```diff
@@ -136,4 +136,4 @@ Detailed API documentation is available for all modules, classes, and functions
 
 [^1]: We also allow for some more general state space models than are described here. For example, the state need not be real (see, for example, the Hidden Markov Model tutorials).
 
-[^2]: As another example of support for general models, we also allow for deterministic dynamical systems (i.e., ODEs).
```
DanWaxman, line 139: Not sure what changed?

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
```diff
@@ -39,6 +39,18 @@
...
+    dtype = None
+    for leaf in jax.tree.leaves(params):
+        if isinstance(leaf, jax.Array) and jnp.issubdtype(leaf.dtype, jnp.inexact):
+            dtype = leaf.dtype
+            break
+    if dtype is None:
+        return arrays
+    return tuple(jnp.asarray(arr, dtype=dtype) for arr in arrays)
+
+
```
DanWaxman: Why is this necessary?

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
```diff
@@ -155,6 +167,9 @@ def _run_linear_kf(
 ) -> PosteriorGSSMFiltered:
     """Run exact continuous-discrete KF (AffineLinearDrift + constant diffusion + LinearGaussianObservation)."""
     params = dsx_to_cdlgssm_params(dynamics)
+    obs_times, obs_values, ctrl_values = _coerce_to_param_dtype(
+        params, obs_times, obs_values, ctrl_values
+    )
```
DanWaxman: What is this doing in practice? Just making integer obs_times float(64), or something? I'd rather avoid type coercion schenanigans.

dynestyx/inference/integrations/cd_dynamax/discrete_filter.py
```diff
@@ -35,6 +35,18 @@
...
+    dtype = None
+    for leaf in jax.tree.leaves(params):
+        if isinstance(leaf, jax.Array) and jnp.issubdtype(leaf.dtype, jnp.inexact):
+            dtype = leaf.dtype
+            break
+    if dtype is None:
+        return arrays
+    return tuple(jnp.asarray(arr, dtype=dtype) for arr in arrays)
+
+
```
DanWaxman: Also, if it is necessary for some reason, refactor it out as a util

dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -16,47 +16,58 @@
...
+    state_dim: int,
+) -> None:
+    """Validate continuous cd-dynamax diffusion compatibility from resolved metadata."""
+    if state_evolution.bm_dim > state_dim:
+        raise ValueError(
+            "Continuous cd-dynamax filters require bm_dim <= state_dim. "
+            f"Got state_dim={state_dim}, bm_dim={state_evolution.bm_dim}."
+        )
+
+
```
DanWaxman: This is very thin, not clear to me why it can't be in other validation.

dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -187,23 +198,24 @@ def dsx_to_cdlgssm_params(dsx_model: DynamicalModel) -> ParamsCDLGSSM:
     - initial_condition is MultivariateNormal
     """
     state_evo = dsx_model.state_evolution
-    if not isinstance(state_evo, ContinuousTimeStateEvolution):
-        raise TypeError("dsx_to_cdlgssm_params requires ContinuousTimeStateEvolution.")
+    if not isinstance(state_evo, StochasticContinuousTimeStateEvolution):
+        raise TypeError(
+            "dsx_to_cdlgssm_params requires StochasticContinuousTimeStateEvolution."
```
DanWaxman: This is not a very useful error for an end user

dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -187,23 +198,24 @@ def dsx_to_cdlgssm_params(dsx_model: DynamicalModel) -> ParamsCDLGSSM:
...
+    # Extract constant L and use resolved Brownian dimension.
     x0 = jnp.zeros(dsx_model.state_dim)
-    L = state_evo.diffusion_coefficient(x0, None, jnp.array(0.0))
-    if state_evo.bm_dim is None:
-        raise ValueError(
-            "state_evolution.bm_dim is not set on ContinuousTimeStateEvolution."
-        )
+    L = state_evo.diffusion.as_matrix(
+        x=x0,
+        u=None,
```
DanWaxman: Why `u = None`? Isn't something like `jnp.zeros(control_dim)` better?

dynestyx/inference/filters.py
```diff
@@ -235,8 +232,10 @@ def _add_log_factors(
             if not isinstance(config, ContinuousTimeConfigs):
                 valid = [c.__name__ for c in ContinuousTimeConfigs]
                 raise ValueError(
-                    f"Invalid filter config: {type(config).__name__}. "
-                    f"Valid config types: {valid}"
+                    "Continuous-time models require a continuous-time filter config. "
+                    "If you want to use a discrete-time filter, nest `Discretizer()` "
+                    "inside `Filter()`. "
```
DanWaxman, line 237: Link to example?

dynestyx/models/core.py
```diff
@@ -151,6 +159,53 @@ def __init__(
 
         # Skip shape validation when inside a numpyro plate context, since
         # batched parameters produce shapes that don't match unbatched expectations.
+
+        def _canonicalize_continuous_state_evolution(
+            *,
+            x_probe: State,
+            u_probe: Control | None,
+            t_probe: Time,
+        ) -> None:
```
DanWaxman: I'm not convinced "canonical" is the right word for what we're doing here. E.g., it's a bit weird to say that `DiagonalDiffusion(jnp.ones(state_dim), state_dim)` is less canonical than `ScalarDiffusion(1.0, state_dim)`.

dynestyx/models/core.py
```diff
@@ -151,6 +159,53 @@ def __init__(
...
+            object.__setattr__(
+                self,
+                "state_evolution",
+                StochasticContinuousTimeStateEvolution(
+                    drift=state_evolution.drift,
+                    potential=state_evolution.potential,
+                    use_negative_gradient=state_evolution.use_negative_gradient,
+                    diffusion=resolved_diffusion,
+                ),
+            )
```
DanWaxman: I don't particularly like `__setattr__` shenanigans

dynestyx/models/core.py
```diff
@@ -151,6 +159,53 @@ def __init__(
...
+                    object.__setattr__(
+                        self,
+                        "state_evolution",
+                        DeterministicContinuousTimeStateEvolution(
+                            drift=state_evolution.drift,
+                            potential=state_evolution.potential,
+                            use_negative_gradient=state_evolution.use_negative_gradient,
+                        ),
+                    )
+                return
```
DanWaxman: Same here

dynestyx/models/core.py
```diff
@@ -159,54 +214,52 @@ def __init__(
...
-                        state_evolution.bm_dim is not None
-                        and inferred_bm_dim != state_evolution.bm_dim
-                    ):
-                        raise ValueError(
-                            "bm_dim does not match inferred diffusion_coefficient output shape. "
-                            f"Got bm_dim={state_evolution.bm_dim}, inferred={inferred_bm_dim}."
-                        )
-                    object.__setattr__(state_evolution, "bm_dim", inferred_bm_dim)
+            # Plate-batched parameters prevent reliable unbatched shape probes, so
+            # we resolve continuous-time diffusion metadata from synthetic vectors.
```
DanWaxman: I don't really get what this means

dynestyx/models/core.py
```diff
@@ -354,6 +403,58 @@ def total_drift(self, x: State, u: Control | None, t: Time) -> dState:
...
+    ):
+        if diffusion.bm_dim is None:
+            raise ValueError(
+                "StochasticContinuousTimeStateEvolution requires diffusion with "
+                "resolved bm_dim."
+            )
+        object.__setattr__(self, "drift", drift)
+        object.__setattr__(self, "potential", potential)
+        object.__setattr__(self, "use_negative_gradient", use_negative_gradient)
+        object.__setattr__(self, "diffusion", diffusion)
```
DanWaxman: Why `__setattr__`?

dynestyx/models/diffusions.py
```diff
@@ -0,0 +1,412 @@
...
+        self,
+        coefficient: DiffusionSpec,
+        bm_dim: int | None = None,
+    ):
+        object.__setattr__(
+            self,
+            "coefficient",
+            coefficient if callable(coefficient) else jnp.asarray(coefficient),
+        )
+        object.__setattr__(self, "bm_dim", None if bm_dim is None else int(bm_dim))
```
DanWaxman: .

dynestyx/models/diffusions.py
```diff
@@ -0,0 +1,412 @@
...
+    def resolve_metadata(
+        self,
+        *,
+        state_dim: int,
+        x_probe: State,
+        u_probe: Control | None,
+        t_probe: Time,
+    ) -> Diffusion:
+        """Validate coefficient shape information and resolve diffusion metadata."""
+        raise NotImplementedError
```
DanWaxman, line 103: Needs better docstring as to what metadata is actually being resolved

dynestyx/models/diffusions.py
```diff
@@ -0,0 +1,412 @@
...
+        if callable(self.coefficient):
+            return None
+        return tuple(int(d) for d in jnp.shape(self.coefficient))
+
+    def _validate_init(self) -> None:
+        if self.bm_dim is not None and int(self.bm_dim) <= 0:
+            raise ValueError(f"bm_dim must be positive. Got bm_dim={self.bm_dim}.")
+
+    def _value_as_matrix(self, value: Array, *, state_dim: int) -> Array:
+        raise NotImplementedError
```
DanWaxman:
```suggestion
        raise NotImplementedError("Please don't construct `Diffusion` directly; instead instantiate one of its subclasses (e.g., `FullDiffusion`, `DiagonalDiffusion`, or `ScalarDiffusion`)")
```

DanWaxman (changes requested):
Didn't get through all the tests, but this looks like a major step! I don't like coercion or `__setattr__` much. Also not sure that we should call a resolved meta-type "canonical."

dynestyx/models/diffusions.py
mattlevine22: put more details in each class

dynestyx/models/core.py
mattlevine22: tried to clarify:

            # In a plate, parameter callables often expect batched parameters, so
            # we resolve diffusion metadata using synthetic per-trajectory probes.

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
mattlevine22: got rid of it.

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
mattlevine22: removed

docs/math_intro.md
mattlevine22: couldn't figure out what changed either.

docs/logo/make_logo.py
mattlevine22: it doesn't work well, didn't fight with it.

docs/api_reference/public/models/core/diffusion.md
mattlevine22: ok tried to improve the docs page.

docs/logo/make_logo.py
DanWaxman: Sure... maybe we should open an issue about it?

dynestyx/inference/integrations/cd_dynamax/utils.py
DanWaxman:
```suggestion
            "dsx_to_cdlgssm_params requires StochasticContinuousTimeStateEvolution. You probably tried to call a continuous-time CD-Dynamax filter with a DeterministicContinuousTimeStateEvolution or DiscreteTimeStateEvolution."
```

dynestyx/inference/integrations/cd_dynamax/utils.py
DanWaxman: Bump

DanWaxman (changes requested):
Thanks! A few other small things that were remaining.

mattlevine22:
Addressed missing `u_probe` in `cd_dynamax/utils.py`; also surfaced an outstanding issue https://github.com/hd-UQ/cd_dynamax/issues/25 which I'll try to address separately.

docs/logo/make_logo.py
mattlevine22: opened #225

dynestyx/models/core.py
```diff
@@ -354,6 +414,58 @@ def total_drift(self, x: State, u: Control | None, t: Time) -> dState:
         return base + grad_term
 
 
+class DeterministicContinuousTimeStateEvolution(ContinuousTimeStateEvolution):
+    """Refined continuous-time state evolution with no diffusion."""
```
DanWaxman:
```suggestion
    """Refined continuous-time state evolution with no diffusion, i.e., describing an ODE."""
```

DanWaxman: Also, we should document these classes better. Even if users aren't going to construct them explicitly, they might show up in type hints, etc.

dynestyx/models/core.py
```diff
@@ -159,54 +215,62 @@ def __init__(
...
+            )
+            u_probe = None if control_dim == 0 else jnp.zeros((control_dim,))
+            t_probe = jnp.array(0.0) if self.t0 is None else self.t0
+            resolved_state_evolution = _refine_continuous_state_evolution(
+                resolved_state_evolution,
+                x_probe=x_probe,
+                u_probe=u_probe,
+                t_probe=t_probe,
+            )
+            self.state_evolution = resolved_state_evolution
```
DanWaxman: I still don't think I understand this block, unfortunately. In particular, I still don't really see how it differs other than (as the earlier comment states) not doing validation. In which case, I think this whole block can be simplified? 

(Sorry I know I'm making this PR a long process, just think having this block be understandable is very important)

DanWaxman (changes requested):
I think probably the last batch of comments :))

dynestyx/models/core.py
mattlevine22: This is the part where we turn the CTSE into Discrete/StochasticCTSE; we use the probes in case it is a callable to check shapes so that we can detect/verify bm_dim as well.

I'm not entirely clear how to make sure I'm playing well with the plates...previously, there was infer_bm_dim code for _inside_plate True and False.

I'll refactor to try to streamline and clarify

DanWaxman: Sure, I guess my point is what's the real difference between here:
https://github.com/BasisResearch/dynestyx/blob/d2f8564948967c384a14d9498d13bd48047f64f9/dynestyx/models/core.py#L226-L239
and the non-plate path:
https://github.com/BasisResearch/dynestyx/blob/d2f8564948967c384a14d9498d13bd48047f64f9/dynestyx/models/core.py#L242-L261
except that one validates and one doesn't?

mattlevine22: Yeah I'm trying to consolidate that now. I think that question could be asked of the current `main` branch as well, right?

DanWaxman: This might've been some sloppy code from before (maybe it was just "early return before validation in plate" -> add in bm_dim inference -> add in probe states -> lots of duplicate code), but we should clean that up now if so, I think.

mattlevine22: agreed, just making sure we are understanding the situation similarly

mattlevine22: should we continue to not validate inside the plate?

mattlevine22: proposing:


```python

        if _inside_plate:
            # Cannot validate shapes with batched parameters; trust the user.
            # Infer observation_dim from observation model if not explicitly provided.
            inferred_obs_dim = _infer_observation_dim_in_plate_context(
                initial_condition=initial_condition,
                observation_model=observation_model,
                inferred_state_dim=inferred_state_dim,
                control_dim=control_dim,
                t_probe=self.t0,
                observation_dim=observation_dim,
            )
        else:
            x_probe, u_probe, t_probe = _make_probes()

            if self.continuous_time:
                _validate_continuous_state_evolution(
                    state_evolution=state_evolution,
                    state_dim=inferred_state_dim,
                    x_probe=x_probe,
                    u_probe=u_probe,
                    t_probe=t_probe,
                )
            else:
                _validate_discrete_state_evolution_output_shape(
                    state_evolution=state_evolution,
                    state_dim=inferred_state_dim,
                    x_probe=x_probe,
                    u_probe=u_probe,
                    t_probe=t_probe,
                )

            obs_dist = observation_model(x_probe, u_probe, t_probe)
            inferred_obs_dim = _infer_vector_dim_from_distribution(
                obs_dist, "observation_model(x, u, t)"
            )
            _validate_observation_dim(observation_dim, inferred_obs_dim)

        self.state_evolution = _resolve_continuous_state_evolution(
            state_evolution,
        )

        self.state_dim = int(inferred_state_dim)
        self.observation_dim = int(inferred_obs_dim)
        self.control_dim = int(control_dim)
        self.categorical_state = bool(inferred_categorical_state)



```

DanWaxman: Doesn't
```
self.state_evolution = _resolve_continuous_state_evolution(
            state_evolution,
        )
```
require the probe states? This looks good to me though, otherwise.

mattlevine22: I refactored to let it use `_make_probe`...will update in a bit.

DanWaxman (approved):
LGTM!

PR #220, opened 2026-05-08
Upgrade Plates For Initial Conditions and Vector-Valued Parameters

### LINKED ISSUE #218: Improve hierarchical tutorials to support multi-dimensional cases
I'm finding it difficult to re-run our hierarchical inference example (even just multiple trajectories) with `state_dim > 1` and `mu_i \in \R^d`. I think this was a key challenge in #216

DanWaxman:
Previously, it was difficult to plate over initial conditions and some vector components (when the number of components matched the state dim). This PR addresses both of those cases. Resolves #218. 

To allow for vector-valued components, under ambiguity, we implement a "whitelist." This whitelist includes things we know will be the same vector shape as the state dimension, e.g., state evolution biases. They are then treated specially, accordingly.

We also now allow for batched i.c.s. This institutes two changes:
- Inside of an initial condition, all vectors inside must be iterated over when deciding vmap shapes.
- Now, to infer state/obs dims, event shapes must be used to avoid ambiguity.

This PR also includes some better documentation for the above cases, including an updated tutorial notebook and better docstring.


docs/tutorials/gentle_intro/08_hierarchical_inference.ipynb
mattlevine22, line 1: The scalar OU example is gone, but the text says "Here we keep the same modeling idea as the scalar OU example, but make everything two-dimensional..."

mattlevine22, line 1: I'd like us to have a single example that works for state_dim=1 and >1. I tried to set `state_dim=1` and changed `A` and `true_params` to be obey `state_dim=1`, but got shape errors.

I'd really like all the sample statements in this `hierarchical_ou_model` to be set up so that it holds that extra state-dim even if scalar---this is how I'd advise people to implement things anyway.

mattlevine22 (changes requested):
Looks good!

Main request is that the tutorial example makes it extremely easy to switch between `state_dim=1` and `state_dim=2` with minimal changes (e.g., intuitively by hand). I wasn't able to convert the 2-d case into 1-d without getting shape errors---can you please look at this?

Other thing is to rectify text in Tutorial: The scalar OU example is gone, but the text says "Here we keep the same modeling idea as the scalar OU example, but make everything two-dimensional..."

DanWaxman:
Not sure what you mean — I did changes by hand and it ran fine. Did you
mean that things like A and sigma are adapted automatically?

mattlevine22:
> Not sure what you mean — I did changes by hand and it ran fine. Did you mean that things like A and sigma are adapted automatically?

No, just by-hand. Can you copy-paste in here what you used? Maybe I forgot something simple.

mattlevine22:
> > Not sure what you mean — I did changes by hand and it ran fine. Did you mean that things like A and sigma are adapted automatically?
> 
> No, just by-hand. Can you copy-paste in here what you used? Maybe I forgot something simple.

Ah scratch that, I fixed it---simple bug on my end. Thanks!

mattlevine22:
Only thing left is to rectify text in Tutorial: The scalar OU example is gone, but the text says "Here we keep the same modeling idea as the scalar OU example, but make everything two-dimensional..."

docs/tutorials/gentle_intro/08_hierarchical_inference.ipynb
DanWaxman: Replaced, thanks!

mattlevine22 (approved):

PR #226, opened 2026-05-15
Add Better Discretizers

DanWaxman:


theorashid:
van Loan? https://github.com/theorashid/cuthbert-models/blob/main/src/cuthbert_models/_discretize.py#L57-L72

I'd heard of it via someone on the pymc statespace team who sent me this https://arxiv.org/abs/2505.18187v1. Apologies if it's in there under a different name

DanWaxman:
> van Loan? https://github.com/theorashid/cuthbert-models/blob/main/src/cuthbert_models/_discretize.py#L57-L72
> 
> I'd heard of it via someone on the pymc statespace team who sent me this https://arxiv.org/abs/2505.18187v1. Apologies if it's in there under a different name

Thanks Theo! It's kinda there now (the "frozen Jacobian" thing is basically take a Jacobian of the drift function, then solve for the matrix exponential; I think the jax compiler optimization would be pretty good here in the case that the drift function is linear). But I agree it should be separate and will make sure to include before marking as ready to review!

mattlevine22:
Can we make sure this also handles discretizing ODEs into mappings between Diracs?

DanWaxman:
Discretizers don’t operate on ODEs currently, though I agree that could be nice as an additional Discretizer (basically as a wrapper to diffrax).

DanWaxman:
Closing this as the branch is somewhat hopelessly outdated. Will put up a newer PR soon.

PR #228, opened 2026-05-17
Move to ty + jaxtyping

### LINKED ISSUE #88: Move to ty for type-checking

### LINKED ISSUE #132: Include `jaxtyping` to better type array shapes
Right now, array shapes are mostly implicit. This would help in making explicit the array shapes we expect (e.g., ndim = 1).

DanWaxman:
Resolves #88. Resolves #132.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,24 @@
+"""Root conftest: compatibility patches applied before test collection."""
+
+import sys
+
+# typeguard 2.13.3 calls evaluate_forwardref(ref, globalns, localns, frozenset())
+# passing recursive_guard positionally.  Python 3.12 made recursive_guard a
+# keyword-only argument, so that call raises TypeError.  We patch the
+# module-level evaluate_forwardref that resolve_forwardref() looks up each
+# call so that the positional frozenset() is moved into the keyword slot.
+if sys.version_info >= (3, 12):
+    import typeguard as _tg
+
+    _orig_evaluate_forwardref = _tg.evaluate_forwardref
+
+    def _patched_evaluate_forwardref(ref, globalns, localns, *args, **kwargs):
+        # args[0] is the old positional recursive_guard (a frozenset)
+        if args and "recursive_guard" not in kwargs:
+            kwargs["recursive_guard"] = args[0]
+            args = args[1:]
+        if "recursive_guard" not in kwargs:
+            kwargs["recursive_guard"] = frozenset()
+        return _orig_evaluate_forwardref(ref, globalns, localns, *args, **kwargs)
+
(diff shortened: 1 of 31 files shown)


dynestyx/inference/filters.py
```diff
@@ -186,9 +189,11 @@ def _add_log_factors(
...
         plate_shapes=(),
-        obs_times: jax.Array | None = None,
-        obs_values: jax.Array | None = None,
-        ctrl_times=None,
+        obs_times: Real[Array, "*obs_time_plate obs_time"] | None = None,
+        obs_values: Real[Array, "*obs_value_plate obs_time observation_dim"]
+        | Real[Array, "*obs_value_plate obs_time"]
+        | None = None,
+        ctrl_times: Real[Array, "*ctrl_time_plate ctrl_time"] | None = None,
         ctrl_values=None,
```
mattlevine22: jaxtype?

dynestyx/inference/filters.py
```diff
@@ -108,7 +111,7 @@ def _add_log_factors(
         plate_shapes=(),
         obs_times=None,
         obs_values=None,
```
mattlevine22: jaxtype?

dynestyx/inference/filters.py
```diff
@@ -69,7 +72,7 @@ def _sample_ds(
...
```
mattlevine22: .

conftest.py
mattlevine22, line 1: can you explain what this is / does

tests/test_utils.py
mattlevine22, line 1: what is the impact of this change?

mattlevine22 (commented):
Looks great!

I have a few small questions, and may have caught some untyped parts?

Overall, I think if the tests pass, I'm happy.

We can double-check the details of the jaxtyping as we use it I suppose...LLM probably got us 95% of the way really fast, no need to spend a bunch of time with a fine-tooth comb on this PR I think.

conftest.py
DanWaxman: This is a file that gets run before tests do. We are now using typeguard, which does some run-time checking of types during tests. But it has a bug that makes it freak out somewhere in Python 3.12, this monkey patches the regressed part so that typeguard still does its job without freaking out at the irrelevant thing.

tests/test_utils.py
DanWaxman: I don't think anything substantive, `ty` just didn't like the `hasattr` type stuff here. We could reinstate the old version with an extra type ignore/cast, if you prefer.

dynestyx/inference/filters.py
DanWaxman: Done!

mattlevine22 (approved):
Awesome thanks!

PR #229, opened 2026-05-18
pin cd-dynamax version to >=0.3.3

mattlevine22:
Prior to v0.3.3 cd-dynamax release, state-dependent diffusion coefficients `L(x, u, t)` silently ignored the `x` argument, instead running `L(None, u, t)`.

This is fixed in v0.3.3: https://github.com/hd-UQ/cd_dynamax/pull/26


DanWaxman (approved):
LGTM so long as tests pass. We should do this for some other libraries, too (e.g., `effectful`); I opened #230 for this.

PR #237, opened 2026-05-22
Improve test coverage for continuous-time hierarchical models + Discretizer 

mattlevine22:
Demonstrating xfailing tests that I would like to work:

Priority level 1:
- seems like Discretizer + Hierarchical is not 
`tests/test_plate_vector_initial_mean_continuous_filters.py`

Priority level 2:
- make reshaping needs consistent and issue helpful errors. For example, why do we need to broadcast initial_mean across M trajectories, but don't need to do this for the initial_covariance? The reason is a bit subtle.
'tests/test_desirable_shared_initial_mean_hierarchical.py'
and
`tests/test_plate_vector_initial_mean_continuous_drift_diffusion.py'


DanWaxman (approved):
I agree that almost all of these should work; thanks!

Issue #213, opened 2026-04-24
Allow For An Interface Without a NumPyro Dependence


DanWaxman:
There may be desires in some settings to avoid a direct NumPyro dependency, dealing with the tracing, etc.

DanWaxman:
Thankfully, this should be rather doable due to some recent refactoring that was required for hierarchical sampling, anyways. 

## Filtering

For example, all filtering integrations now come with a pure JAX implementation; for example, in cuthbert integrations:

https://github.com/BasisResearch/dynestyx/blob/c70d726fe98e423d02bc8267ded8434f49e1f578/dynestyx/inference/integrations/cuthbert/discrete.py#L163-L177

The corresponding NumPyro code that surrounds that simply runs the `compute_cuthbert_filter` path, then records some `numpyro.deterministic` sites + `numpyro.factor` statements:

https://github.com/BasisResearch/dynestyx/blob/c70d726fe98e423d02bc8267ded8434f49e1f578/dynestyx/inference/integrations/cuthbert/discrete.py#L238-L287

Similar results hold for CD-Dynamax integrations.

## Simulators

Simulators are a bit more messy, but I think still doable. 

SDE simulator would be pretty straightforward. It mostly comprises a pure-JAX `solve_sde` call + a rather simple `numpyro`-based `emit_observations` call:

https://github.com/BasisResearch/dynestyx/blob/c70d726fe98e423d02bc8267ded8434f49e1f578/dynestyx/simulators.py#L819-L851

This `_emit_observations` is really not doing anything special `NumPyro`-wise:

https://github.com/BasisResearch/dynestyx/blob/c70d726fe98e423d02bc8267ded8434f49e1f578/dynestyx/simulators.py#L591-L630

ODE simulation is a similar story.

The more painful story will be discrete-time simulation. To be honest, this is kind of a mess right now, from trying to support multiple different features (perfect observations, hierarchical simulation, batched simulation, and normal simulation). It contains a pure-JAX scan:
https://github.com/BasisResearch/dynestyx/blob/c70d726fe98e423d02bc8267ded8434f49e1f578/dynestyx/simulators.py#L1012-L1047

But it also contains a lot of other stuff... I think this will be a major pain point that we have to deal with.

## Other Caveats

[...]

PR #238, opened 2026-05-26
dsx.infer for filtering and smoothing without numpyro

theorashid:
This is for https://github.com/BasisResearch/dynestyx/issues/213.

So far it is just for `Filter` and `Smoother`. Simulator will have to come.

Basic idea is that Filter/Smoother no longer call `numpyro.factor` inline, it happens in `dsx.sample`. `test_filters` and `test_smoothers` still pass, happy days.

```python
# non-numpyro path: returns InferResult directly (no side effects)
with Filter(filter_config=KFConfig(...)):
    result = dsx.infer("f", dynamics, obs_times=t, obs_values=y)
loss = -result.marginal_loglik

# numpyro path (unchanged from before):
with Filter(filter_config=KFConfig(...)):
    dsx.sample("f", dynamics, obs_times=t, obs_values=y)
```

In the background, `sample` just calls `infer` and then registers sites. `dsx.infer` returns `InferResult` which carries `marginal_loglik`, `states`, `dists`, and a private `_register_numpyro_sites` callback.

```python
def sample(...):
    result = infer(...)
    if isinstance(result, InferResult) and result._register_numpyro_sites is not None:
        result._register_numpyro_sites(name)
    return result
```

The main work can be seen in the `test_*_standalone.py`, which I based from my cuthbert-models repo. This is how I would see it being used.

```python
def test_infer_optax_mle():
    """Use dsx.infer + optax to do MLE without numpyro."""
    obs_times, obs_values = _make_data()

    def neg_loglik(alpha):
        dynamics = _make_lti_dynamics(alpha)
        with Filter(filter_config=KFConfig(filter_source="cuthbert")):
            result = dsx.infer(
                "f", dynamics, obs_times=obs_times, obs_values=obs_values
            )
        return -result.marginal_loglik

    optimizer = optax.adam(1e-2)
    alpha = jnp.array(0.3)
    opt_state = optimizer.init(alpha)

    initial_loss = neg_loglik(alpha)
    grad_fn = jax.grad(neg_loglik)

    for _ in range(20):
        grads = grad_fn(alpha)
        updates, opt_state = optimizer.update(grads, opt_state)
        alpha = optax.apply_updates(alpha, updates)

    final_loss = neg_loglik(alpha)
    assert final_loss < initial_loss
```

## how these changes were made

Of course, this was largely done by burning tokens and handholding so that it matches the design that I (and then with feedback from both of you) wanted. I am not as familiar with the internals of the library, so if there are other places beyond the diff that you think these changes might affect, let me know where to look – I was relying a bit on existing tests not breaking.

## smaller design things

I largely tried to keep everything that was there before to reduce the size of the refactor. But:

- I did make `BaseLogFactorAdder` (and the `Filter` equivalent) an `ABC` because I think it made sense.
- Plate, as before, does not register per-field sites.
- Maybe we should rename `_sample_intp` to `_infer_intp`.
- `InferResult` a `__call__` shim to satisfy `FunctionOfTime` protocol (needed because some model functions `return dsx.sample(...)`). Effectful uses the `@defop` return annotation to decide the type of `fwd()` returns. `InferResult.__call__ `raises `NotImplementedError` to satisfy the protocol  — we can't change it without reworking `Simulator`.


theorashid:
On the testing suite, sometimes `tests/test_hierarchical_simulator_discretizer_smokes.py`, `tests/test_science/test_discrete_time_l63_mcmc.py` are flaky. `tests/test_science/test_hmm.py::test_mcmc_inference` breaks for TypeError: only 0-dimensional arrays can be converted to Python scalars. `test_science` is pretty slow in general

DanWaxman:
Thanks Theo! This seems directionally about right, I like the implementation strategy!

> In the background, sample just calls infer and then registers sites. dsx.infer returns InferResult which carries marginal_loglik, states, dists, and a private _register_numpyro_sites callback.

That makes sense! It will be a tiny bit tricky to get working with Simulators, I think, but that's probably okay. One tricky part is you're allowed to stack simulators with filters, i.e., `with Simulator(), Filter(): dsx.sample(...)`; this allows one to sample from the filtering/posterior predictive. We'll need to pass the corresponding `InferResult` and append the necessary information.

> I am not as familiar with the internals of the library, so if there are other places beyond the diff that you think these changes might affect, let me know where to look

I don't have any off the top of my head, besides the aforementioned interaction with Simulators. The other main test will be running all the notebooks in the documentation from scratch and making sure the results are qualitatively similar, but I think that can come in a bit. 

> I did make BaseLogFactorAdder (and the Filter equivalent) an ABC because I think it made sense.

Agreed, thanks!

> Maybe we should rename _sample_intp to _infer_intp.

Sure, I more-or-less agree.

> sometimes tests/test_hierarchical_simulator_discretizer_smokes.py, tests/test_science/test_discrete_time_l63_mcmc.py are flaky

I'm surprised the simulator/discretizer smokes are flaky, I haven't run into that before... but it's not super surprising to me that the discrete_time_l63_mcmc is flaky. I think those were written before we made EnKF the discrete-time default.

> test_science is pretty slow in general

Right... I think the test_science suite has somewhat fallen into disuse, and our tests in general are a bit of a mess (though I think with decent coverage -- just disorganized). It's been sitting on the backlog for a bit. I wouldn't worry too much about particularly slow test_science tests as long as docs in the notebooks are looking okay.

---

At the risk of getting ahead of myself on a draft PR, I think it makes sense to set up a staging branch for this. Then, we can try to land this PR in the staging branch; worry about Simulators and its various interactions afterwards; then worry about the documentation lift that this implies.

theorashid:
>I think the test_science suite has somewhat fallen into disuse

Right, now I've seen the workflows I can see they just run the tests ignoring `test_science`.

I can give `Simulator` a go if you want it in this PR, but I'm wary of keeping the size small so it's easy for you to review – up to you.

DanWaxman:
> I can give `Simulator` a go if you want it in this PR, but I'm wary of keeping the size small so it's easy for you to review – up to you.

I think it makes sense to keep the PRs small, but also want to minimize the amount of half-working features on the upstream. So I've changed the base branch to `dsx-infer-staging`, where we can work bit-by-bit in implementing `dsx.infer(...)`.

From that perspective, feel free to mark as ready to review whenever you feel it's ready and I'll take a close look :) thanks again!!

theorashid:
Just renamed `_sample_intp` to `_infer_intp`.

I think this is a good point. Smaller change, tests passing, not tooooo many files to check over. Then I'll use any feedback before doing the `Simulator` refactor.

mattlevine22:
1. Yeah, you can ignore `test_science` for now. 
2. Did you say `tests/test_hierarchical_simulator_discretizer_smokes.py` is having issues? That one should be a solid test, but it looks like everything passed in (green checks on your previous commits).
3. I just ran tutorials 04 (discrete-time filter + simulator roll-out) and 06 (SDE filter + simulator roll-out), and both worked and look right locally, so that is a good sign!
4. I think I prefer `dsx.condition`, as I worry that `infer` will sound "all-powerful" to some users trying to do parameter estimation. Infer will sound a bit weird when we use it for Simulator rollouts, but it is not CRAZY to say that a simulator rollout is "conditioned", even if only on `obs_times`.

dynestyx/inference/integrations/cd_dynamax/discrete_filter.py
```diff
@@ -167,8 +133,15 @@ def run_discrete_filter(
...
-) -> list[dist.Distribution]:
-    """Run discrete-time filter via cd-dynamax (KF, EKF, UKF)."""
+) -> tuple[jax.Array, object, list[dist.Distribution]]:
+    """Run discrete-time filter via cd-dynamax (KF, EKF, UKF).
+
+    Pure computation — no numpyro side-effects. Callers are responsible for
+    registering numpyro.factor / numpyro.deterministic if needed.
+
+    Returns:
+        tuple: (marginal_loglik, posterior, filtered_dists).
```
DanWaxman: Should match the other return docstring block

https://github.com/theorashid/dynestyx/blob/536ead8c9a99b71ad6ab7dd26b4c02b4cab767cd/dynestyx/inference/integrations/cd_dynamax/continuous_filter.py#L212-L218

dynestyx/inference/integrations/cd_dynamax/discrete_smoother.py
```diff
@@ -106,8 +74,15 @@ def run_discrete_smoother(
...
-) -> list[dist.Distribution]:
-    """Run discrete-time smoother via cd-dynamax (KF, EKF, UKF)."""
+) -> tuple[jax.Array, object, list[dist.Distribution]]:
+    """Run discrete-time smoother via cd-dynamax (KF, EKF, UKF).
+
+    Pure computation — no numpyro side-effects. Callers are responsible for
+    registering numpyro.factor / numpyro.deterministic if needed.
+
+    Returns:
+        tuple: (marginal_loglik, posterior, smoothed_dists).
```
DanWaxman: Same here (should match other more detailed docstring)

dynestyx/inference/filters.py
```diff
@@ -237,7 +249,19 @@ def _add_log_factors(
...
+            key = config.crn_seed
+        else:
+            import warnings  # noqa: PLC0415
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                try:
+                    key = numpyro.prng_key()
+                except Exception:
+                    key = None
```
DanWaxman: Not a huge fan of the bare try/except

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -260,15 +255,18 @@ def run_discrete_filter(
...
+    registering numpyro.factor / numpyro.deterministic if needed.
+
     Returns:
-        list[dist.Distribution]: Filtered state distributions at each obs time.
+        tuple: (marginal_loglik, raw_states, filtered_dists).
     """
     obs_len = int(obs_values.shape[0])
     if obs_len == 0:
-        return []
+        return jnp.array(0.0), None, []
```
DanWaxman:
```suggestion
        return None, None, []
```

I think the regestration guard is actually on the MLL:

https://github.com/theorashid/dynestyx/blob/536ead8c9a99b71ad6ab7dd26b4c02b4cab767cd/dynestyx/inference/filters.py#L348-L350

dynestyx/inference/integrations/cuthbert/discrete_smoother.py
```diff
@@ -351,11 +257,15 @@ def run_discrete_smoother(
...
+) -> tuple[jax.Array, object, list[dist.Distribution]]:
+    """Run discrete-time smoother via cuthbert.
+
+    Returns:
+        tuple: (marginal_loglik, raw_states, smoothed_dists).
+    """
     t1 = int(obs_values.shape[0])
     if t1 == 0:
-        return []
+        return jnp.array(0.0), None, []
```
DanWaxman:
```suggestion
        return None, None, []
```

As before, I think the guarding is on the MLL:

https://github.com/theorashid/dynestyx/blob/536ead8c9a99b71ad6ab7dd26b4c02b4cab767cd/dynestyx/inference/smoothers.py#L333-L335

dynestyx/inference/numpyro_sites.py
```diff
@@ -0,0 +1,361 @@
+"""NumPyro site registration for filter and smoother outputs.
+
+All numpyro.factor and numpyro.deterministic calls live here,
+keeping the integration backends (cuthbert, cd-dynamax) numpyro-free.
```
DanWaxman: except for numpyro.distributions

dynestyx/handlers.py
```diff
@@ -106,10 +78,50 @@ def sample(
...
+    obs_values: Real[Array, "*obs_value_plate obs_time observation_dim"]
+    | Real[Array, "*obs_value_plate obs_time"]
+    | None = None,
+    ctrl_times: Real[Array, "*ctrl_time_plate ctrl_time"] | None = None,
+    ctrl_values: Real[Array, "*ctrl_value_plate ctrl_time control_dim"]
+    | Real[Array, "*ctrl_value_plate ctrl_time"]
+    | None = None,
+    predict_times: Real[Array, "*predict_time_plate predict_time"] | None = None,
+    **kwargs,
+):
```
DanWaxman: After some discussion with @mattlevine22, I agree that `condition` may be clearer than `infer`

DanWaxman (changes requested):
This is pretty close on my end! A few things:
- After talking with Matt, I agree that `dsx.condition(...)` is likely to be clearer than `dsx.infer(...)`. `condition` is much closer to the statistical "work" that `infer` is actually doing.
- Some inconsistencies in docstring should be fixed
- The guards during registering results are on `marginal_loglikelihood` being `None`, but currently, empty filters return an increment of `0.0`. It's not super clear what this means for registering sites as a result. 
  - Would also be nice to have a test for empty filters

theorashid:
I've still kept `InferResult`. `ConditionResult` sounds weird, it sounds like the result of some `if-else` logic rather than what we do here.

DanWaxman:
Thanks, will take another look tomorrow! 

> I've still kept `InferResult`. `ConditionResult` sounds weird, it sounds like the result of some `if-else` logic rather than what we do here.

Hmm, I agree... Maybe `ConditioningResult`? I don't think this has to be a blocking point.

mattlevine22:
Agree not to let it block...I'd probably do ConditionedResult fwiw

dynestyx/inference/filters.py
```diff
@@ -237,7 +249,16 @@ def _add_log_factors(
...
+        # Resolve PRNG key: use explicit seed from config, fall back to numpyro
+        # context (inside a seeded model), or None (deterministic filters don't need one).
+        if config.crn_seed is not None:
+            key = config.crn_seed
+        else:
+            import warnings  # noqa: PLC0415
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                key = numpyro.prng_key()  # returns None outside seed handler
```
DanWaxman, line 268: This also needn't be blocking right now, but this is a legitimate shortcoming; we don't really have a way to deal with non-CRN filters without using the numpyro `seed` handler. This is non-trivial to fix (though one option is to basically replicate the `numpyro.prng_key()` function/`seed` handler, which are not very complex and are a rather straightforward way to implement global random seeds).

DanWaxman (approved):
I think this looks very reasonable. Certainly good enough to merge into a staging branch, for me. There is an actual limitation of this API right now (we don't have a great way to deal with non-CRN filters without invoking the `numpyro.seed` handler), though. That should be dealt with at some point.

theorashid:
`ConditionedResult` sounds good. I changed it – start as we mean to go on

mattlevine22:
@theorashid would you mind adding a notebook in the `deep_dive` section of docs that gives a demonstration of these new features (and points out any sharp edges the code may have) akin to our Tutorial notebooks....basically, the demo you wish you had when you wanted to use dynestyx at first :) I find this complimentary to PR Summaries / tests / function documentation, and will help us understand how things look at a high-level as this overall PR push moves along. Note that we can clean up this sort of documentation after we merge this current PR, but good to have a minimal version sooner rather than later.

theorashid:
Yeah definitely. I was planning on contributing notebooks in subsequent PRs, specifically porting some dynamax SSMs to dynestyx, and then rewriting one of these using in the canonical parametrisation with your help.

But for this PR here, what would be useful? Perhaps an "outer loops" type tutorial, where I show several ways of doing inference on the hyperparameters: optim via optax with `condition`, SVI/MCMC with `sample` (anything else?). What do you think?

mattlevine22:
What do you think about replicating these tutorials without NumPyro PPL:

- https://basisresearch.github.io/dynestyx/stable/tutorials/gentle_intro/03_filtering_mll/
- https://basisresearch.github.io/dynestyx/stable/tutorials/gentle_intro/05_svi/ (just do optax MLE, not SVI)

If you can reproduce all the figures in those 2 notebooks, I think we are doing good and ready to move on!!

theorashid:
Nice. Where would they live? Directly overwrite the current tutorials using `condition`? Or add another section in those tutorials to show both with/without numpyro? 

I'll have to get to this next week. I'm 1 day into 3 full days of team meetings (sigh) so no fun stuff for me this week :-(

DanWaxman:
> Nice. Where would they live? Directly overwrite the current tutorials using `condition`? Or add another section in those tutorials to show both with/without numpyro?

Sorry forgot to respond to these! I think for now it makes sense to just make a copy of those notebooks. We can figure out what to do with them a bit later. 

In the meantime, we've merged a few things into main, and this now has merge conflicts; I'll try to sort these out now and make a PR to `theorashid:refactor/decouple-numpyro-filters-smoothers`. While I'm at it, I'll also make a quick pass at these notebooks.

theorashid:
Yeah sorry both, work has been heavier than usual these past 2 weeks. I've blocked time on Monday to do this though, but if you wanted to fix the conflicts before (or try the notebooks) that would be dreamy.

DanWaxman:
Update: @theorashid I opened https://github.com/theorashid/dynestyx/pull/1 for this. @mattlevine22 would be great if you could also take a look at the two `_no_numpyro` notebooks you requested.

DanWaxman:
> Yeah sorry both, work has been heavier than usual these past 2 weeks. I've blocked time on Monday to do this though, but if you wanted to fix the conflicts before (or try the notebooks) that would be dreamy.

No worries at all, grateful you're taking the time to contribute in the first place :)

mattlevine22:
Yes will look over today. Why did some of the numpyro tutorials change?

DanWaxman:
Lint error is unrelated, will merge to staging branch after tests pass. Thanks Theo!

theorashid:
Once merged, shall I do Simulator refactor next week?

DanWaxman:
> Once merged, shall I do Simulator refactor next week?

That would be great!!

PR #239, opened 2026-05-28
Missingness support for Simulators

mattlevine22:
**Summary**

This PR adds missing-observation support for simulator-based inference with non-Dirac observation models, plus a new tutorial showing how to use `Simulator` + MCMC with missing data.

On the implementation side, `DiscreteTimeSimulator` and `ODESimulator` now support `obs_values` containing `NaN`s under an index-preserving interpretation: `obs_values[k]` always refers to latent index `k`, so missing observations do not remove latent states from the model. Full-row missingness is supported for all non-Dirac observation models by simply skipping the observation log-probability at those indices. Partial missingness is supported when the observation model has structure that can be marginalized cleanly, specifically:
- `MultivariateNormal` observations
- factorized `Independent(..., 1)` observations

To keep the simulator code clean, the missing-data logic is factored into a shared internal scorer layer in [dynestyx/internal/observation_missingness.py](/Users/levinema/Projects/research/dynestyx/dynestyx/internal/observation_missingness.py). When `NaN`s are present, the simulator switches from `obs=...` sample sites to `numpyro.factor(...)`-based masked observation scoring while preserving the existing no-missing fast paths.

The PR also adds a focused test suite under [tests/missingness/](/Users/levinema/Projects/research/dynestyx/tests/missingness/) covering:
- scorer math,
- discrete simulator behavior,
- ODE simulator behavior,
- hierarchical/plated behavior.

Finally, it adds a new gentle-intro tutorial notebook, [11b_missing_observations_simulator_mcmc.ipynb](/Users/levinema/Projects/research/dynestyx/docs/tutorials/gentle_intro/11b_missing_observations_simulator_mcmc.ipynb), which explains the full-vs-partial missingness distinction for simulator-based inference, compares the Gaussian case against KF smoothing, and illustrates additional non-Gaussian observation families supported by the simulator missingness path.

Questions / notes:
- do we really want the missingness utils in `dynestyx/internal/observation_missingness.py`?


mattlevine22:
@copilot does the most recent commit address your comments/concerns?

docs/tutorials/gentle_intro/00_index.ipynb
```diff
@@ -1,60 +1,38 @@
...
... and Bayesian parameter estimation).\n\n9. **[Part 8: Hierarchical / mixed-effect inference](../08_hierarchical_inference/)** \u2014 `plate` for trajectory-level parameters; hierarchical dynamical models and inference.\n\n10. **[Part 9: Discrete-time smoothing](../09_discrete_smoothing/)** \u2014 `Smoother` for discrete-time models; smoothed state summaries and future-only posterior rollout.\n\n11. **[Part 10: Continuous-time smoothing](../10_continuous_smoothing/)** \u2014 Continuous-discrete smoothing; smoothed state summaries and future-only posterior rollout.\n\n12. **[Part 11: Missing observations with Filters and Smoothers](../11_missing_observations/)** \u2014 Whole-row and partial missingness with the cuthbert Kalman filter/smoother path.\n\n13. **[Part 11b: Missing observations with `Simulator` + MCMC](../11b_missing_observations_simulator_mcmc/)** \u2014 Joint posterior inference over parameters and latent states under Gaussian missing-data simulator conditioning.\n\n14. **[Part 12: Hierarchical modeling patterns](../12_hierarchical_modeling_patterns/)** \u2014 Plates, shared vs. local parameters, nested hierarchies, controls, missing data, and current backend caveats.\n"
```
DanWaxman: Does 12 actually exist?

docs/tutorials/gentle_intro/11b_missing_observations_simulator_mcmc.ipynb
```diff
@@ -0,0 +1,932 @@
...
...p2I4P6XehHRt5R484QefAAfvTOBK+LL0dLYCK+vMDgi6HX71k1obix0tlcYF2Hw8s3CSDf3t/uhjJpJT+zaBfMdP0ONpMUti96GplADHCoLzNEMVGQeeNpG4ENvA87cKm7KpmuvEUKS4ZEXEcmW3sgeD+3EFa/kkTMZgLNPKxzoi89B9vYPIWnWoCmUx8FNDkTjGgTDpZEDOrUWtz1ixpd+G0V776FJnEpANBlHvNEHh8aGS5rPgerKC6E6ZxtUG9dBtXQRlqzbhnfW3ohVxiWIPf4YVH/+J/Kf/j8syuaF2DRVipXnMrFoxYcgCTt90T7s8+3DgcABOOODsEVlMXiiQeR0l2ofg9EIUySNVDomzHO7va3Q+EIwOmpnLGXu+ZcPYuO6FSJiaTzI24gEJqqOQlCKG3W46L1qWL64WYhS5SKHFjTV4abXXyLS2Ir7SaHpV15YMCw+WijyqqWprmy0DEXR6HTaOeHTNF76RSgVRsqoQj4eg2qWDVwpQsbv64NBVglT2+uuuRyWEZFszyhteOx8BxadXpr+dcGay/DyGhUs+7qgjIrMoY4/pRjTK4l95CG0ckV11zd5GnnCcaSS4UnlyFM1K/rR2Gzievcl/ciq1XBvvbXEVNeoNcKhd8Cf9sNryCDjcYsQ6dHL8nfuQ1CThklvr8qX7sFHnsIv//A3hCJRtDQ1iAit4qCvHOtWr4SZqrmNg8lRi7C7D4nkkXt0Kh4WXnRqw5HvLsnE8dH789jUXXr9ve6fXnz/xzns8hQ8tuj5dPd9DwsBeKroLQ5kPV64IgOIeAdgjWWAkd6FdJ+LyyJKhKp4MScgVC3SdxBKThKTc415adKpUAazFXI0glSq8iAtlUrhnHPOwZUXnItQcBBdzmBF7zOCnvN6Ww2ygy74QoVJwVDch1B3K6yW2jFic7XQvlH1vLAUgaLVoMbSCHMgDsiFSQMlGhWWAo+/+PKklqtkZCh/uBuKPyjS7368vBP/fWUA+z99LfCDLwEffRfwH++A6gM3iyIDVEG2CHlU2ax1CEY88IzyVTtWUHQHpRam8hJCuZgQECnay2GsEZFuo/tBI58jlZhMJPpENDfWib7ERHgDYTz27JEIiPFIJSXUO+xwJrzojPbDnQxASsl48flD0OY1IvJprkPbSOOCkBSrGB1O0Uxkfm4dx8CdlmPRGoXgRP8eSHrQE
```
DanWaxman: It's true that we should use particle filtering/smoothing here, but that's because EKF is going to do really bad with Laplace observations.

dynestyx/simulators.py
```diff
@@ -961,6 +990,64 @@ class DiscreteTimeSimulator(BaseSimulator):
...
+        scorer = build_missing_observation_scorer(
+            observation_model=dynamics.observation_model,
+            missing_data=missing_data,
+        )
+        state_transition = cast(DiscreteStateTransition, dynamics.state_evolution)
+
+        with numpyro.plate(f"{name}_n_simulations", 1):
+            x_prev_site: Real[Array, " state_dim"] | Real[Array, ""] = numpyro.sample(  # type: ignore
+                f"{name}_x_0", dynamics.initial_condition
+            )
```
DanWaxman, line 908: Why size 1?

dynestyx/simulators.py
```diff
@@ -961,6 +990,64 @@ class DiscreteTimeSimulator(BaseSimulator):
...
+        x_prev = x_prev_site[0]
+
+        u_0 = _get_val_or_None(ctrl_values, 0)
+        numpyro.factor(
+            f"{name}_y_0_lp",
+            scorer.score_step(x=x_prev, u=u_0, t=times[0], t_idx=0),
```
DanWaxman: I'd like to avoid "score", it's somewhat overloaded for us

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,194 @@
...
+    cov = obs_dist.covariance_matrix
+    mask_outer = mask_f[:, None] * mask_f[None, :]
+    masked_cov = cov * mask_outer + jnp.diag(1.0 - mask_f)
+
+    chol = jnp.linalg.cholesky(masked_cov)
+    whitened = jsp.linalg.solve_triangular(chol, residual, lower=True)
+    quad = jnp.dot(whitened, whitened)
+    logdet = 2.0 * jnp.sum(jnp.log(jnp.diag(chol)))
+    n_obs = jnp.sum(mask_f)
+    return -0.5 * (quad + logdet + n_obs * LOG_2PI)
```
DanWaxman: Is there not a nice numpyro way to do this with a MaskedDistribution?

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def _is_factorized_independent_distribution(obs_dist: dist.Distribution) -> bool:
+    return isinstance(obs_dist, dist.Independent) and (
+        obs_dist.reinterpreted_batch_ndims == 1
+    )
```
DanWaxman: Very thin helper

DanWaxman (changes requested):
I think this is directionally right, but I have a few concerns:

1. There's a lot of new complexity; `observation_missingness` has a dataclass, `Protocol`, datahelper,  scorer class, and factory, along with some math. This is a lot of complexity when it feels like just the data helper and scorer class should suffice. 
2. Simulators is getting quite long. It may be out of scope for this PR, but I do think it's time to break up that class if we're adding these new features. 
3. I don't really want totally separate pathways for the missing data unless it's totally necessary (and I feel it probably isn't). The `_simulate_missing_scan` is doing almost the same thing as the normal simulation loop, with an abstracted `scorer` (which I would like to change the name of); we should just use that abstracted scorer directly, always, if we're going to already add the new complexity.

docs/tutorials/gentle_intro/00_index.ipynb
mattlevine22: removed, thanks

dynestyx/simulators.py
mattlevine22: renamed things to `log_potential`

dynestyx/internal/observation_missingness.py
mattlevine22: removed

dynestyx/internal/observation_missingness.py
mattlevine22: not really, MaskedDistribution is for when you have a boolean array on batch-dims

docs/tutorials/gentle_intro/11b_missing_observations_simulator_mcmc.ipynb
mattlevine22: I changed to:

If we want to perform Filtering/Smoothing as before, we would need to use a variant that can handle these non-Gaussian observations (e.g. PF/PS or EKF/EKS); however, as noted in the previous tutorial, these methods are not yet set up to automatically support missing data. The EnKF, while configured to automatically support missing data, is not set up to support non-Gaussian observations.

Nevertheless, the `MCMC + Simulator()` strategy can handle all forms of missingness (partial and full) under non-Gaussian observation models as long as the observation model is independent per observation dimension, which is the case here.

dynestyx/simulators.py
mattlevine22: this appears in `main` as well...I think it is to keep a leading axis for consistency w.r.t. when we do multiple rollout simulations

dynestyx/observation_missingness.py
DanWaxman, line 1: I'm still not 100% convinced a new `internal` directory is the right place for this, though I'm okay with it being a new file.

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,163 @@
...
+from dynestyx.models.checkers import _make_probe_state
+from dynestyx.models.core import DynamicalModel
+
+LOG_2PI = jnp.log(2.0 * jnp.pi)
+
+
+def _masked_multivariate_normal_log_prob(
+    obs_dist: dist.MultivariateNormal,
+    y: Array,
+    obs_mask: Array,
```
DanWaxman: jaxtyping

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,163 @@
...
+    """Evaluate conditioned observation log-potentials and preserve NaNs in outputs."""
+
+    dynamics: DynamicalModel
+    obs_values: Array
+    distribution_mode: Literal[
+        "uninitialized", "masked", "multivariate_normal", "independent"
+    ] = dataclasses.field(init=False, default="uninitialized")
+    safe_obs: Array = dataclasses.field(init=False)
+    obs_mask: Array = dataclasses.field(init=False)
+    row_has_any_observed: Array = dataclasses.field(init=False)
```
DanWaxman: jaxtyping

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,163 @@
...
+    whitened = jsp.linalg.solve_triangular(chol, residual, lower=True)
+    quad = jnp.dot(whitened, whitened)
+    logdet = 2.0 * jnp.sum(jnp.log(jnp.diag(chol)))
+    n_obs = jnp.sum(mask_f)
+    return -0.5 * (quad + logdet + n_obs * LOG_2PI)
+
+
+@dataclasses.dataclass
+class ObservationLogPotential:
+    """Evaluate conditioned observation log-potentials and preserve NaNs in outputs."""
```
DanWaxman: We should comment a bit more on what this class is doing

dynestyx/simulators.py
DanWaxman: Fair enough

dynestyx/internal/observation_missingness.py
```diff
@@ -0,0 +1,163 @@
...
+        t_probe = jnp.array(0.0) if self.dynamics.t0 is None else self.dynamics.t0
+        self._configure_from_distribution(
+            self.dynamics.observation_model(x=x_probe, u=u_probe, t=t_probe)
+        )
+
+    def _is_scalar_like_row(self) -> bool:
+        return self.obs_values.ndim <= 1 or self.obs_values.shape[-1] == 1
+
+    def _configure_from_distribution(self, obs_dist: dist.Distribution) -> None:
+        """Choose the log-potential evaluation mode once before scanning in time."""
```
DanWaxman: We should be more explicit somewhere that we don't support changing distribution type if you have partial observations.

dynestyx/simulators.py
```diff
@@ -857,6 +883,69 @@ class DiscreteTimeSimulator(BaseSimulator):
...
+            numpyro.factor(
+                f"{name}_y_{t_idx + 1}_lp",
+                log_potential.log_potential_step(
+                    x=x_t,
+                    u=u_next,
+                    t=t_next,
+                    t_idx=t_idx + 1,
+                ),
+            )
+            y_t = log_potential.observation_step(t_idx + 1)
```
DanWaxman: Should we be adding a deterministic site for y_t?

dynestyx/simulators.py
```diff
@@ -1154,8 +1267,9 @@ def _simulate(
             dynamics: A `DynamicalModel` whose `state_evolution` is a
                 `DeterministicContinuousTimeStateEvolution`.
             obs_times: Times at which to save the latent state and emit observations.
-            obs_values: Optional observation array. If provided, observation sites are
-                conditioned via `obs=obs_values[i]`.
+            obs_values: Optional observation array. If provided for a non-Dirac
+                observation model, conditioning is handled via per-step
+                log-potential factors.
```
DanWaxman: Need to explain log-potential factors if we use it in docstring

DanWaxman (changes requested):
Looks close to me! A few remaining things:
- we should try to use jaxtyping as much as possible (though I acknowledge our coverage of this in simulators isn't great, most of the annotations went into filter stuff; maybe this should be a separate issue)
- if we're going to adapt the FK nomenclature for the potential (which I like!), we should either not use it in docstrings, or explain what it means in docstrings
- I think we're missing a numpyro.deterministic site for y?

dynestyx/observation_missingness.py
mattlevine22: moved to a standalone file

dynestyx/internal/observation_missingness.py
mattlevine22: added

dynestyx/internal/observation_missingness.py
mattlevine22: has a longer description

dynestyx/internal/observation_missingness.py
mattlevine22: now we error if things change in-loop (just via a cheap check), and adding it to high-level DiscreteTime/ODESimulator docstring

dynestyx/simulators.py
mattlevine22: done

DanWaxman (approved):
This looks good to me! Thanks for all the work on it!!

PR #240, opened 2026-05-28
Plating Fixes After New Test Patterns

DanWaxman:
#237 includes a few new tests with desired use patterns for plating. This PR identifies and fixes a few different problems to close the gap:

## Initial condition handling

Initial conditions were a pain point that kept causing problems. This is because of various oddities in how numpyro distributions work; for one, the `batch_shape` is a static property of the pytree, so slicing the initial condition doesn't update the batch shape accordingly. Second was that `MultivariateNormal` often stores a leading singleton dimension for no apparent reason, which broadcasts shapes unintentionally. 

To address this, we now branch according to if an initial condition is batched or not. If it is, we flatten the corresponding pytree, broadcast all vectors to the full plate-padded shape, and vmap afterwards. 

## Parameters under closure

An antipattern that was causing issues in tests was that a plated parameter appeared under a Python closure. For example, something like 
```
with dsx.plate("M", 10):
    alpha = numpyro.sample(...)
    def drift(x, u, t): foo(alpha)
```
This uses the original 10-dimensional alpha in the closure of `drift`, and is therefore not sliced properly in vmaps later on.

The correct pattern is to define such things as equinox modules, with properly-defined parameters. This fixes the failing tests in `tests/test_plate_vector_initial_mean_continuous_drift_diffusion.py`. 

## Discretizer with bias

Another issue was that bias variables caused issues under `Discretizer`. This is because `bias` functions as a "whitelisted" variable under plating, such that it is properly understood to be `state_dim` and not a plating dimension. But under discretizer, the original `bias` term is stored in a corresponding `cte` object. These are now included in the whitelist so they're properly vmapped over.

## Other Changes

I've also updated the tutorial notebook with the discretizer, and a "sharp bits" section explaining some choices as above. I also added some additional tests, and fixed the shape error in one of the xfailed tests.


mattlevine22 (changes requested):
this looks great to me, thanks Dan!
- all xfailed tests now pass
- I didn't notice any slowdowns / performance hits when running the hierarchical notebook
- The improved hierarchical tutorial does a nice job of showing how (and why) to make an equinox class drift. 

One request---in the sharp-edges section of the tutorial, I think (2) `to_event(1)` would warrant a more explicit DO and DON'T (I like that 1 does this). I think (3) covers it with "do MVN not Normal", but if you think there is a clear DO and DON'T, I suggest adding that as well.


I also asked copilot to review---maybe it will find a bug or something (though you probably already did some LLM checks).

mattlevine22 (approved):
Looks great, thanks!!

PR #241, opened 2026-05-29
Fix quickstart link in README

SamDuffield:
Updated quickstart link in README to point to stable version.

The "Versioned documentation" link was also broken although I wasn't sure what this was now supposed to link to so I removed it (this may not be the best solution).


DanWaxman (approved):
Thanks Sam!! I'm not too worried about the versioning link, this solution looks good to me.

PR #242, opened 2026-05-29
Strengthen NaN checks in smoke tests and surface nested-plate NaN bug

mattlevine22:
This PR strengthens our test suite’s NaN detection by adding two shared helpers in [tests/test_utils.py](/Users/levinema/Projects/research/dynestyx/tests/test_utils.py): `assert_tree_all_finite(...)` for arrays / pytrees, and `assert_trace_sites_exist_and_field_all_finite(...)` for NumPyro traces. The trace helper now checks both that sites exist and that their selected field is finite, so a lot of trace-based tests got simpler and more consistent.

I also added a couple of focused tests for the new trace helper itself, covering both failure modes: a missing trace site and a non-finite trace value. Then I applied the helper to the highest-priority simulator / filter / smoother smoke tests that were previously mostly shape-only.

In the process, we found a real hidden NaN issue in the nested-plate discrete path: nested plate discrete conditioning currently leaves NaNs in `f_states`. That behavior is now captured with a strict `xfail`, and the related nested Dirac rollout case is also marked `xfail` since it hits the same underlying nested-plate NaN problem.

### DIFF
--- a/tests/test_smoothers.py
+++ b/tests/test_smoothers.py
@@ -39,6 +39,10 @@ from tests.models import (
     discrete_time_lti_simplified_model,
     jumpy_controls_model_ode,
 )
+from tests.test_utils import (
+    assert_trace_sites_exist_and_field_all_finite,
+    assert_tree_all_finite,
+)
 
 
 def _gen_obs_discrete():
@@ -117,10 +121,24 @@ def test_compute_cuthbert_smoother_returns_observation_aligned_states(
     )
 
     assert jnp.ndim(marginal_loglik) == 0
+    assert_tree_all_finite(
+        {"marginal_loglik": marginal_loglik}, where="smoother output"
+    )
     if isinstance(smoother_config, PFSmootherConfig):
+        assert_tree_all_finite(
+            {
+                "particles": states.particles,
+                "log_weights": states.log_weights,
+            },
+            where="PF smoother states",
+        )
         assert states.particles.shape[0] == len(obs_times)
         assert states.log_weights.shape[0] == len(obs_times)
     else:
+        assert_tree_all_finite(
+            {"mean": states.mean, "chol_cov": states.chol_cov},
+            where="Gaussian smoother states",
+        )
         assert states.mean.shape[0] == len(obs_times)
         assert states.chol_cov.shape[0] == len(obs_times)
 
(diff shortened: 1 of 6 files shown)


tests/test_utils.py
```diff
@@ -9,11 +9,105 @@
...
+        if site_name not in tr:
+            raise AssertionError(f"{where}: site {site_name!r} not found in trace")
+
+        site = tr[site_name]
+        if field not in site:
+            raise AssertionError(
+                f"{where}: site {site_name!r} does not contain field {field!r}"
+            )
+
+        selected[site_name] = site[field]
```
DanWaxman, line 74: One thing that might be nice is to not greedily return the first missing/infinite/nan site, but collect all missing/inifnite/nan sites and return a single error at the end.

tests/test_smoothers.py
```diff
@@ -511,6 +558,13 @@ def test_smoother_plate_batched_continuous_initial_condition_ct_rollout():
     assert tr["f_marginal_loglik"]["value"].shape == (3,)
     assert tr["f_predicted_times"]["value"].shape == (3, 1, len(predict_times))
     assert tr["f_predicted_states"]["value"].shape == (3, 1, len(predict_times), 2)
+    assert_trace_sites_exist_and_field_all_finite(
+        tr,
+        "f_marginal_loglik",
+        "f_predicted_times",
+        "f_predicted_states",
+        where="continuous plate smoother trace",
+    )
```
DanWaxman, line 564: Shape asserts after existence?

tests/test_smoothers.py
```diff
@@ -492,6 +532,13 @@ def test_smoother_plate_batched_continuous_initial_condition_discretized_rollout
     assert tr["f_marginal_loglik"]["value"].shape == (3,)
     assert tr["f_predicted_times"]["value"].shape == (3, 1, len(predict_times))
     assert tr["f_predicted_states"]["value"].shape == (3, 1, len(predict_times), 2)
+    assert_trace_sites_exist_and_field_all_finite(
+        tr,
+        "f_marginal_loglik",
+        "f_predicted_times",
+        "f_predicted_states",
+        where="discretized plate smoother trace",
+    )
```
DanWaxman, line 538: Same

DanWaxman (changes requested):

DanWaxman (approved):
This looks good for now. One thing for the future, since we plan on overhauling our tests at some point soon anyways; some of these hand-rolled test utilities are implemented in [`chex`](https://chex.readthedocs.io/en/latest/); we should probably use `chex` directly in the future.

PR #244, opened 2026-06-01
Fix Frozen Diffusion Coefficient Array; Properly Plate Diffusion Coefficients

DanWaxman:
This PR fixes two things:
- An issue with an accidentally-static diffusion field in `StochasticContinuousTimeStateEvolution` (thanks @rfl-urbaniak for pointing this out!)
- A resulting bug with shared diffusion coefficients under plating, that the preceding bug accidentally masked.

For the latter, the issue is that plate dimensions with scalar/vector quantities can cause ambiguities with plate dimensions, since we still don't annotate event dimensions. This now treats diffusions with constant coefficients as opaque leaves when plating or indexing, and uses an annotated event dimensions property to break the ambiguity. 

Note that this leaves two places with exceptions (vector-shaped biases, and constant diffusions). This makes me sad. It makes sense to, some time soon, consider a type of annotated meta-data system (this is essentially a type of named tensor dimension) that properly annotates event dimensions. But, this PR should at least make the built-in types work correctly.

### DIFF
--- a/dynestyx/inference/plate_utils.py
+++ b/dynestyx/inference/plate_utils.py
@@ -3,9 +3,12 @@ import jax.numpy as jnp
 import numpyro
 from jaxtyping import Array, Shaped
 
+from dynestyx.models import Diffusion
 from dynestyx.utils import (
     _array_has_plate_dims,
+    _diffusion_coefficient_is_plate_batched,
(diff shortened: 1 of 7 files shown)


mattlevine22 (approved):
Looks good, thanks @DanWaxman !

PR #246, opened 2026-06-02
Pin minimal viable versions of dependencies

### LINKED ISSUE #230: Pin Minimal Viable Versions of Libraries
We have a few libraries that should have minimum versions specified -- e.g., I'm sure older `effectful` versions would not work, and we are sticking to `mkdocs<2` for now.

DanWaxman:
Resolves #230. Specify minimum versions for libraries that previously lacked them, and be explicit about diffrax and numpy dependencies.

- effectful: pin >=0.2.0 (first release exposing effectful.ops.types.NotHandled,  which handlers.py imports; 0.1.x/0.0.x cannot import dynestyx)
- diffrax: declare >=0.6.2 (used by solvers/ and simulators.py; cd-dynamax pins it to ==0.6.2, so this is the minimum resolved version)
- numpy: declare >=2.0
- optax: add to dev group as >=0.2.0 (imported by the SVI smoke tests)
- mkdocs: add the <2 cap
- remove a duplicate nbconvert entry in the dev group

Minimum versions all pass test suite.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "effectful",
+    "effectful>=0.2.0",
     "cuthbert>=0.0.10",
     "cuthbertlib>=0.0.10",
     "cd-dynamax>=0.3.3",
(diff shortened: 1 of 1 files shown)


mattlevine22 (approved):
LGTM

Issue #114, opened 2026-02-25
Add support for linear time-varying systems


mattlevine22:
- Requires more plumbing to Dynamax / CD-Dynamax

mattlevine22:
Completed by #250

PR #250, opened 2026-06-10
Add support for time-varying linear systems

DanWaxman:
Previously, Kalman filters/smoothers only support time-invariant systems; to this end, both LinearGaussianStateEvolution and LinearGaussianObservation were time-invariant. This is a problem when trying to include proper discretization schemes for CD linear systems, where time-varying linear systems naturally arise with non-uniform sampling (cf. #226). This PR adds time-varying support for linear Gaussian models.

Main changes are to (i) the LinearGaussianStateEvolution interface, (ii) the LinearGaussianObservation interface, (iii) the KF/RTS Smoothing cuthbert integrations.

Changes (i) and (ii) are quite similar; they both allow for parameters to depend on `(t_now,t_next)`/`t` or be constant, in a potentially mixed way. They are resolved via a `params_at` function, so that data about the system may still be extracted for KF/RTS use.

This makes the changes for filtering and smoothing integrations somewhat minimal. They just go from accessing the info to using `evo.params_at(t_now, t_next)`.

For `dynamax` (which I recall being somewhat fragile regarding LTV systems), we supply a `is_time_invariant` flag, and throw errors accordingly. We make additional guards at the integration stage later on as well.


dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -183,6 +183,36 @@ def _initialize_model_params(model: Any, **raw_kwargs: Any) -> Any:
...
+    component: Any,
+    field_names: tuple[str, ...],
+    *,
+    where: str,
+    suggestion: str = (
+        "Use fixed arrays, or choose the cuthbert backend (e.g. "
+        "KFConfig(filter_source='cuthbert') / "
+        "KFSmootherConfig(filter_source='cuthbert')), which supports "
+        "time-varying linear-Gaussian models."
+    ),
```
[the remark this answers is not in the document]
mattlevine22: Nope, the current text is correct

dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -232,6 +262,16 @@ def dsx_to_cdlgssm_params(dsx_model: DynamicalModel) -> ParamsCDLGSSM:
     obs = dsx_model.observation_model
     if not isinstance(obs, LinearGaussianObservation):
         raise TypeError("dsx_to_cdlgssm_params requires LinearGaussianObservation.")
+    _require_constant_linear_gaussian_fields(
```
mattlevine22, line 265: I'm fine with this---we could probably support Dynamax's time-varying KF. It expects `A.shape =  (times, state, state)`, so would just have to pre-build the extra dimension. Doesn't seem worth the hassle here.

dynestyx/inference/integrations/cd_dynamax/utils.py
```diff
@@ -346,15 +386,29 @@ def dsx_to_cd_dynamax(
...
+                "which supports time-varying linear-Gaussian models."
+            ),
+        )
+
+        # The guard above ensures the parameters are constant arrays.
+        H_matrix = cast(jnp.ndarray, obs.H)
+        D_matrix = None if obs.D is None else cast(jnp.ndarray, obs.D)
 
         def emission_function(x, u, t):
             if x.ndim > 1:
```
mattlevine22, line 405: ugh these bug me---not for this PR, but eventually (when we fully jaxtype the simulators), I'm excited to get rid of these checks for scalar cases (instead lifting them up to a vector earlier on)

DanWaxman: Haha, agreed... though I think this is doing something a bit different: `obs.D` can be a `callable` or `jnp.ndarray`, but is about to be used in a way that requires it to be a `jnp.ndarray`. But the type-checker isn't smart enough to know we just tested to make sure `obs.D` is a matrix and doesn't have to worry about the `callable` possibility, so we have to tell it.

dynestyx/inference/integrations/cuthbert/discrete_smoother.py
```diff
@@ -19,6 +19,7 @@
     CuthbertInputs,
     _config_to_filter_kwargs,
     _drop_cuthbert_dummy_step,
+    _kalman_dynamics_params_builder,
```
mattlevine22, line 22: surprised this file doesn't also import `_kalman_observation_params_builder` as well

DanWaxman: RTS smoother doesn't need observations

DanWaxman: (other than what's stored from the forward pass, obviously)

dynestyx/models/state_evolution.py
```diff
@@ -15,6 +15,24 @@
 from dynestyx.models.core import DiscreteTimeStateEvolution
 
 
+class LinearGaussianParams(NamedTuple):
```
mattlevine22, line 18: this is basically the same as the Observation version of this class, other than names. Makes me wonder if our LinearGaussian classes should not end in `Observation` or `StateEvolution`...they can all just have something like `state2out_matrix, control2out_matrix, bias, cov`

mattlevine22 (approved):
Looks like things work here and the PR accomplishes its mission, so approving!

One question I'll let you think about and/or leave for later.
- Should we unify LinearGaussianObservation and LinearGaussianStateEvolution into a single class? or at least just have them shadow a single underlying class? ...same question w.r.t.  the new `Params` version of these classes. I realized these are all literally the same other than field names.

dynestyx/inference/integrations/cuthbert/discrete_smoother.py
mattlevine22: ah yes I see, thanks!

DanWaxman:
> Should we unify LinearGaussianObservation and LinearGaussianStateEvolution into a single class? or at least just have them shadow a single underlying class? ...same question w.r.t.  the new `Params` version of these classes. I realized these are all literally the same other than field names.

They're not with callables -- state evolution has signature `(x, t_now, t_next) ->` ... while observation has `(x, t) -> ...`.

mattlevine22:
Closes #114

PR #253, opened 2026-06-12
Agent skill for paper implementation with dynestyx

EddieCunningham:
Users of dynestyx include practitioners and researchers who want to implement research papers relevant to their work.  Doing this by hand takes time even when dynestyx handles all of the modeling and inference, because the user still has to write the application-dependent code, and the modeling and inference are only one part of that.  AI agents can speed this process up, but they often need structured guidance to implement the requirements correctly.  This agent skill can provide that structure.

This skill is called `implement-with-dynestyx`, and lives in a new folder called "agent_skills" containing a SKILL.md that summarizes what the agent should do, and a few other files containing examples of usage patterns.  This folder will also contain another skill called [grill-with-docs](https://github.com/mattpocock/skills/tree/main/skills/engineering/grill-with-docs) that the agent will invoke to grill the user on requirements.

The user would prompt an LLM with something along the lines of "Use the agent skill at agent_skills/implement-with-dynestyx to implement this paper ...", and then the agent will be told to do the following 5 things:

1.  Find, or ask the user for, easily readable context for the paper, such as latex source or a reference codebase.
2. Understand the details of the paper.
3. Interactively grill the user on their requirements.
4. Determine how dynestyx can be used for the implementation.
5. Build, test and run the deliverable.

It's not obvious right now what the best way to evaluate this skill is, so I think its worth discussing what exactly this skill should help with and how to test its utility.


mattlevine22:
"""It's not obvious right now what the best way to evaluate this skill is, so I think its worth discussing what exactly this skill should help with and how to test its utility."""


100% agree, and curious what your preliminary thoughts are?

From looking at the notebooks you linked to via email (I'm not sure I looked at the right stuff, so please correct me / my impressions), it seemed like the agents wrote largely reasonable code, but that the ultimate results were quite different from the papers they were prompted with (at least in the epidemiological case, seemed like `omega` was close to 0 unlike in the paper).

One level of evaluating the skill is "does it help us write reasonable dynestyx code we would have written by hand"...I think that is mostly true, but to be fair, codex seems to do this well on my laptop too.

The next-level, which I don't think we've quite hit (and may just be an overall challenge with current models at a given price point), is successful experimentation itself. 
--> Did you view any of the produced notebooks to be close to "scientifically successful / helpful"?

Concretely, I noticed that some of the notebooks ran SVI and the elbo was still decreasing substantially at the end of the optimization. On one hand, I think dynestyx could have better optimization tooling. On the other hand, I would have hoped an agent saw this lack of convergence and played with things until optimization ran more to completion.

Similarly, the simplest plot in the epidemiology paper was the Marginal Log Likelihood plot w.r.t. `omega`...maybe I missed it, but, I don't think any of the skill/planning agents actually made this plot.

So...do you think there are some easier targets/requests that we can ask of an agent?

EddieCunningham:
I agree with your assessment, what you saw were the first outputs from the agent and to improve them you'd need to iterate.  I guess we need to understand how to do this iteration, what target we want to hit, and how to measure how useful the skill was in hitting this target in comparison to not using it at all.

> The next-level, which I don't think we've quite hit (and may just be an overall challenge with current models at a given price point), is successful experimentation itself.

This seems like a good target to try to reach.   Maybe the user should specify a specific experiment that they want to do and the paper that is passed in can be treated as context.  The skill would still contain dynestyx best practices in reference files, like how to build models and run inference, but the main skill could focus on extracting the precise requirements of the experiment from the user.  We could also add a "grill the user + iterate" step to the end of the skill so that the process continues until the user is happy with the result, but I'm not sure if that's necessary because the user will do this anyways.

Another option could be to leave the design of an experiment completely to a user and instead write skills so that a user is able to prompt at a higher level of abstraction.  For example, we could write a skill that contains the best practices for how to build models and run inference with dynestyx.  I've personally found this to be useful with my projects.  For example, these days if I want to write code for a generative model, I would write a prompt that looks something like this "Build and train a diffusion model for my data using [/linsdex](https://github.com/EddieCunningham/linsdex/blob/main/.cursor/skills/linsdex/SKILL.md) in a self contained file.  The model should parameterize the drift and use a brownian bridge as the base SDE.  /grill-with-docs".

On top of these choices, I'm not sure what the best way to evaluate how good a skill is, compared to not using it at all.  We can measure things like how long it takes to implement an experiment from start to finish, number of development iterations, and total cost in tokens, but this is time consuming because getting a single data point requires a human to go through the entire development process each time.  Also, there are confounding factors like what agent was used, what system prompts the user has in their local settings, what kind of harness is built around the agent, etc.  I'm not sure what level of rigor makes sense for evaluating a skill like this.  What do you think?

DanWaxman:
FWIW, I have typically had success in implementing models with pretty vague instructions (something like "write a 2d OU process with coupling constant $\rho$, with independent Poisson emissions of rate $\exp(x_1 + x_2)$ and $\exp(x_1 - x_2)"). Lots of the models in tutorials are more or less written this way (usually with a pointer like "read the other docs in @docs/ to understand idiomatic usage of dynestyx").* So I suspect there is something about "getting lost" in a paper and/or the data wrangling.

Maybe to this end it is worth it to interact with the user to a shared understanding of the mathematical model being implement, then "translate" this into dynestyx code?

\* Indeed, I prompted ChatGPT 5.5 Extra High in Codex with the prompt 
> Read the docs in @docs/ to gain an understanding of idiomatic dynestyx usage. Then write a model describing a 2d OU process with coupling constant $\rho$, with independent Poisson emissions of rate $\exp(x_1 + x_2)$ and $\exp(x_1 - x_2)$. Plot simulated data and a posterior over $\rho$ via SVI using a particle filter. Write this in a notebook docs/prompting_illustration.ipynb.

and got a [reasonable one-shot answer](https://gist.github.com/DanWaxman/6f3692143a5b92851d573395aacc84a4).

EddieCunningham:
I re-did the skill to focus on getting the model right.  It does this by creating a pdf that describes the model and focuses on grilling the user on the different parts of it.  I tested it on [this paper](https://www.biorxiv.org/content/10.64898/2026.06.01.729394v1).  Here is the [pdf](https://github.com/user-attachments/files/29109759/lake_trout_ssm.pdf) that was created, the code for the [model](https://github.com/user-attachments/files/29109762/model.py), and code to recreate figure 2 analytically [recreate_figure2.py](https://github.com/user-attachments/files/29109765/recreate_figure2.py) and using simulations from the dynestyx model [recreate_figure2_simulated.py](https://github.com/user-attachments/files/29109763/recreate_figure2_simulated.py).  Here are the plots they generated:

figure2_recreation

figure2_recreation_simulated

DanWaxman:
Hey Eddie! Sorry for taking so long to get back to this, had some travel etc. for the end of June.

I guess, in general, I'm not super convinced by the efficacy of LLMs at implementing the "right" model in the paper, and I don't feel great about packaging skills I'm not confident in. Do you think it would still be useful to package a simpler skill for basic usage? E.g., explain some of the idiomatic use of `dynestyx`, sharp bits, capabilities, and some examples?

EddieCunningham:
Hi Dan, no worries.  I agree that it seems like LLMs aren't good enough yet to extract the right model from a paper without a lot of human intervention - this turned out to be way more difficult than I originally expected!  However, they do seem very capable of writing dynestyx code after a prompt or two so I honestly don't think a skill explaining how to use dynestyx will add much over just asking it to explore the codebase.  In any case, this was interesting to try out!

DanWaxman:
Makes sense, thanks so much for trying it out! I agree it was super interesting (and makes me less existential 😆)

PR #254, opened 2026-06-12
Update cmap usage

DanWaxman:
In the newest matplotlib, plt.cm.get_cmap was removed; we should use plt.get_cmap now.


mattlevine22 (approved):
Looks fine, my local HMM tutorial still plots correctly on this branch.

PR #255, opened 2026-06-12
Staging Branch for Numpyro-Free Usage

### LINKED ISSUE #138: Add support for missing data in discrete-time models

### LINKED ISSUE #213: Allow For An Interface Without a NumPyro Dependence

### LINKED ISSUE #258: Simulator improvements
Simulators.py needs some love!
Possibly dependent on #155 (decoupling from numpyro, work in progress) but could happen before.

1. Overall, it is too long and needs a refactor.
2. Naming should be improved---let's reserve Simulator for data-generation, and pick something else for inference (e.g., JointInference? or could even just be the default interpretation of every model conditioned on data.)
3. Latent state log-probs can be computed via either jax.vmap or `scan`, but currently we mostly do scan. See #138 [comment](https://github.com/BasisResearch/dynestyx/issues/138#issuecomment-4681890641). Should unify this with DiracIdentity and missingness.
4. Should have an "online" simulator that enables active control-feedback loops (i.e. filter, execute a controller, then continue simulating).

Will become much more rich with #226 (more discretizers).

### LINKED ISSUE #271: Numpyro-divorce w/ Simulator Refactor
Complete #255 (and complete #266 along the way).

DanWaxman:
Staging branch for numpyro-free usage.

---

This PR provides a pathway for NumPyro-free usage, alongside other Simulator-based changes that happened during this refactor. At a high level:
- Pure-Jax computations under a `filter` and `smoother` are completely isolated. It is now possible to run a filter or smoother without entering the NumPyro effect handler stack using `dsx.condition(...)`, which only calls these pure-Jax parts.
- This leads to `dsx.sample(...)` now working by (i) calling `dsx.condition(...)`, which registers some callbacks, then (ii) calling these callbacks to create NumPyro sites.
- `Simulator`s are now generative-only, and pure-Jax. This simplifies their presentation significantly, gets rid of some naming people repeatedly found confusing, and allows joint (theta, x) inference to be a bit fancier with maintainable code.
  - `Simulator` remains a handler for `dsx.sample()`
  - `dsx.simulate(...)` is its new JAX-free api.
- The rollout-based inference component of `Simulator` is now a `LatentPathBuilder`. This is a bit of a fancy component, and only supports usage with NumPyro.
  - It supports arbitrary missingness patterns via an "augment" pattern: this strategy creates "placeholder" sites, fills any missing sites, and scores after. This requires a new custom class which inherits from `ImproperNormal` but carries some "correct"-ish sampling semantics.
  - It also supports analytical marginalization of `Independent` and `MultivariateNormal` distributions.
  - `dsx.log_prob(...)` is the new JAX-free api for this work, but is much less fancy---it simply scores a dynamical model and a state-reconstruction against the provided data.
- Lots of documentation changes to match these updates. 

- [x] Filtering support (#238)
- [x] Smoothing support (#238) 
- [x] Simulator support (#266)
- [x] Docs changes (#280)


DanWaxman:
@theorashid I set up this draft PR as staging for the numpyro-free stuff. My proposed workflow would be to open new PRs targeting this branch (`dsx-infer-staging`) for the remaining items (namely, support for simulators and the corresponding documentation changes).

theorashid:
So I'm less familiar with `Simulator` and its use cases. Let's discuss it first before I implement it.

The rough aim is to extend what I did to Filter/Smoother so that any of the Simulator-backed models (discrete-time, ODE, SDE) can be fit with optim (or other) instead of numpyro

Similar to what we had to filter, I'd like to use `condition` as

```python
def neg_loglik(params):
    dynamics = build_model(params)
    with Simulator():
        result = dsx.condition("f", dynamics, obs_times=t, obs_values=y)
    return -result.marginal_loglik   # differentiable -> optax

# vs the unchanged numpyro path:
with Simulator():
    dsx.sample("f", dynamics, obs_times=t, obs_values=y)
```    

I think `Simulator` is harder than `Filter` because it uses numpyro as the actual compute substrate (`numpyro.sample("x_0", ...)` to draw latent states, `numpyro.factor` for the observation log-density), so a standalone path needs more than deferring recording.

ODE seems easiest because the latent path is deterministic (no random draws). SDE _should_ be okay too because it is only forward-simulating (and raises on `obs_times`/`obs_values`).

Some qs for you:

1. Am I right that the latent-draw coupling does not block any sort of inference? I think `numpyro.sample("x_0", ...)` is only for simulation / prior-predictive so isn't needed during parameter inference. The conditioning log-density is just `numpyro.factor(log_prob)`, which I can compute and defer like I did for `Filter`.

2. What should `ConditionedResult` track for a `Simulator`? Filter returns `marginal_loglik`, `states`, `dists`. Should the times/states/observations trajectories go into `ConditionedResult` as a new field or trajectories dict, or should we just leave that for numpyro deterministic sites only with condition just returning the loglik

3. Do we want standalone forward simulation outside of a numpyro model (to generate synthetic data)? I don't see why not, the numpyro pieces are just `numpyro.prng_key()` (explicit `crn_seed` key, like `Filter`) and the `numpyro.sample(...)` draws (`dist.sample(key) + lax.scan`) to make the `n_simulations > 1` branch work. The current numpyro path names sites `(x_0, y_0, x_1, …)` and relies on `Predictive` for the `n_simulations` plate. Could standalone return trajectories in ConditionedResult instead of named sites? (And then the dsx.sample wrapper re-attaches numpyro sites). I'm also wary of swapping `nscan` for `lax.scan` and the plate-suspension logic (`_suspend_numpyro_plate_frames`). Anything that I should be careful about — e.g. semantics that only hold inside a numpyro trace?

4. Posterior rollout (filter-then-predict). In a numpyro model, `with Simulator(), Filter(): dsx.sample(...)` chains automatically — Filter `fwd()`s `filtered_dists` up to the Simulator, which rolls past the obs window. For standalone condition, do you prefer

```python
# single-handler. user wires the handoff
with Filter(...):
    filtered = dsx.condition("f", dynamics, obs_times=t, obs_values=y)
with Simulator():
    predicted = dsx.condition("f", roll_from(filtered.dists[-1]), predict_times=future_t)

# chained. condition threads one handler's result into the next like the numpyro path:
with Simulator(), Filter(...):
    result = dsx.condition("f", dynamics, obs_times=t, obs_values=y, predict_times=future_t)
```

I think I prefer the explicit style of the single-handler.

DanWaxman:
Hey Theo! I think this will be related to #258. Maybe it makes sense ot find a time to Zoom and talk these things through? Otherwise, I'll try to meet with Matt tomorrow and make some decisions/suggestions about these things.

theorashid:
See what you and Matt think and whether this or the Simulator refactor should come first, but happy to have a call if it isn't obvious. I can do most mornings 9-10:30 pacific.

mattlevine22:
Okay so here is the summary from my discussion with Dan:
1. We should split "simulators" into a) `simulators` that only generate / simulate data and b) methods that allow us to work explicitly with latent states `x_k` (formerly the role of Simulators and now in need of a new name, e.g. `LatentStateBuilder`? PathUnroller? etc..) 
2. We should also refactor how those things work, both to accommodate some needed functionality in #258, as well as support numpyro-trace-free usage of `simulators` and `LatentStateBuilder`. These usages include two new numpyro-trace-free functions:
- `dsx.log_prob(dyn_model, latents=[x1,...xk], ... obs/ctrl=..., chunk_size=...)` 
    - compute `log p(y1...yk | x1...xk, ... )`
    - `chunk_size` allows this to be computed as a scan of vmaps in cases of big data
- `dsx.simulate(dyn_model, predict_times=..., discretizer=...)` 
    - generate data from HMM/SSM/ODE/SDE

Questions for @theorashid: Do the above 2 numpyro-trace-free functions look like things you might want? What are we missing?

Proposal: Once we agree on a direction, I'm happy to help move it forward; @theorashid for this part of the PR, it would be great if you could help us make sure we expose the right numpyro-trace-free stuff and build a simple doc/demo/tutorial for how you would recommend using it.


# Details #
## Simulator for Data Generation

`Simulator` should become a strictly generative model: it should roll out an SSM and/or solve the ODE/SDE at specified times.

### Proposed behavior

- `Simulator` should call a NumPyro-trace-free subroutine to do the actual rollout/solve.
- This can support both non-`diffrax` and `diffrax` implementations, including `UnsafeBrownianTree`, since we do not need adjoints/gradients here.
- If used via `dsx.sample(...)`, it should additionally attach sites to the NumPyro trace, matching current practice.
- I recommend creating `SDESimulatorConfig` and `ODESimulatorConfig` instead of exposing the user to different simulator classes; this aligns better with our patterns for Filters/Smoothers and also makes things easier when we want aliases that don't have `with` statements, like `dsx.simulate(...)`

### Usage without NumPyro

```python
with Simulator(SimConfig): # don't need a config if using discretizer since it is just discretetime.
    with Discretizer():  # optionally
        simulation_result = dsx.condition(
            dynamical_model,
            ...,
            predict_times=...,
        )
```

### Possible alias

```python
simulation_result = dsx.simulate(
    dynamical_model,
    ...,
    predict_times=...,
    discretizer=..., #optionally,
    config=
)
```

### Usage with NumPyro

```python
with Simulator(SimConfig): # don't need a config if using discretizer since it is just discretetime.
    with Discretizer():  # optionally
        dsx.sample(
            dynamical_model,
            ...,
            predict_times=...,
        )
```

---

## Simulator for Inference

The other role of old-Simulator was to support "joint state + parameter inference".
Basically, it created latent sites `x_k` that aligned with observations and put them into the numpyro trace (for later inference). 
We want to make this more modular and usable without a numpyro trace.


### Joint state+parameter inference without a NumPyro trace

```python

def loss(latents, params):
    dynamical_model = build_model(params)
    return dsx.log_prob(
        dynamical_model,
        latents=latents, #[x1, ..., xk]
        ...,
        obs_values=..., #[y1, ..., yk]
        chunk_size=...,
    )

# run optimziation for argmin_{latents, params} loss(latents, params)
```

### Joint state+parameter inference  with NumPyro

This would look similar to current usage, but under a new name  and with a refactored implementation that uses `.log_prob(...)` under-the-hood. Dan and I have struggled a bit with vocab in this section, so opinions are very welcome re: naming!

```python
def dyn_model(...):
    ...
    dsx.sample("f", DynamicalModel(...), obs/ctrl/predict/etc)

with InterpretationX(chunk_size=...):  # e.g. LatentStateBuilder?
    nuts_kernel = NUTS(dyn_model)
    mcmc = MCMC(nuts_kernel, num_warmup=100, num_samples=100)
    mcmc.run(jr.PRNGKey(1), obs/ctrl...)
```

---

## Workflow for `InterpretationX`

1. Create the needed latent states in the NumPyro trace (i.e. `x_k` for all `k`).
2. Call `log_prob(...)`, which computes `p(y_k | x_k)` for all `k`.

Given all `x_k`, this computation is fully factorizable across `k`.

### Performance strategy

- For small datasets, we would likely want to `vmap`.
- For large datasets, we would likely want to chunk and `lax.scan` over smaller `vmap`s.

---

## Missingness Notes

`log_prob(...)` can use the existing strategies for ignoring or marginalizing missing observations.

Special case: `DiracIdentityObservation` with zero observation noise and fully observed state.

In that case:

- Only create latent states where there are `NaN`s.
- For all other times, set `x_k = y_k`.
- Mostly score transitions.
- Only score observations of `y_{k+1}` if `y_k` had missing values.

theorashid:
Yea so I think those functions you proposed look like what I'd want. All I want is 1) to be able to run inference without a numpyro trace (optax/MLE) for any model (SSM/ODE/SDE...) and 2) clear separation of concerns (`numpyro` sites as an optional layer) + composability (via `fwd()`). The functions match the separation I was reaching for in my [cuthbert-models](https://github.com/theorashid/cuthbert-models/blob/main/tests/test_nonlinear_continuous.py#L202-L212) prototype, where the trace-free op (`infer`) returned plain `Posterior` and numpyro registration was a separate optional layer:

```python
# yours
with Simulator(SimConfig):
    with Discretizer():  # optional
        simulation_result = dsx.condition(dynamical_model, ..., predict_times=...)

simulation_result = dsx.simulate(dynamical_model, ..., predict_times=..., discretizer=..., config=...)

# numpyro version, unchanged
with Simulator(SimConfig):
    dsx.sample(dynamical_model, ..., predict_times=...)

# mine (cuthbert-models)
# no numpyro
with Filter(EKF(...)), Discretizer(EulerMaruyama()):
    posterior = infer(model, emissions, obs_times=t)
loss = -posterior.marginal_loglik

# mine (cuthbert-models)
# NumpyroTrace on to register sites
with Filter(EKF(...)), Discretizer(EulerMaruyama()), NumpyroTrace("f"):
    infer(model, y, obs_times=obs_times)

# but now we have `_register_numpyro_sites` within `sample` so no need for NumpyroTrace
```

`dsx.log_prob` is new and is not something I considered – a scoring function for joint log-density of an explicit latent path plus observations, but seems useful for joint state+param inference:

```python
# I guess something like this? Is that what you were thinking?
latents0 = dsx.simulate(model0, predict_times=t, key=k).states

def loss(latents, params):
    dynamical_model = build_model(params)
    return dsx.log_prob(dynamical_model, latents=latents, obs_values=..., chunk_size=...)
# and then optimise over both latents and params?
```

And it is distinct from `condition` (takes latents in rather than producing/marginalizing them). Seems like a good idea to separate out the `DiscreteTimeSimulator` logic into this. There might be a better name for it? `numpyro.dist.log_prob` exists. Maybe because `condition` does marginal, we could do `log_joint`, maybe even `path_log_joint`?

As for splitting the work, like I say I'm less familiar `Simulator` – you might have the clearest ideas here. Maybe it's best for you to lead and I review, but I can take a stab at it if you want? I think if you wanted to refactor `Simulator` (#258) to mainline directly, that should go first so the ground is stable (and then we separate out the numpyro trace parts). But if they're too coupled we could go straight into this branch.

mattlevine22:
Ok cool; at a high-level, I guess you are primarily interested in `simulate` for generating synthetic data / rolling out forecasts?

Sometimes, joint inference is easier and faster than marginalizing with a filter, so we want to support both (and this is currently done w/ Simulator). 

For naming, maybe `dsx.path_log_prob`? I worry that "joint" can be too confusing (does it include parameter priors?)

As far as workflow, I do think things are coupled enough that it will be best to put these changes directly here.

I'm happy to take the crack at this; I will be hiking in Banff this week, but back online on Friday.

mattlevine22:
@theorashid Can you take a look at my merge with main #265 ? Once that is in, I'll fire up the simulator refactor and we can get this PR back on track. I'd really like us to have this merged by end of next week if possible (it is a great improvement to the repo, but kinda blocks further development). How is your timing looking? Seems like you mostly built the stuff you want/need from dynestyx in the filtering part of this PR?

theorashid:
I'll give the branch a review by end of Friday definitely, but will try to do it sooner :-)

mattlevine22 (approved):
Great job everyone, I think this:
- streamlined Dynestyx usage a lot (including numpyro-free, but also more than that!)
- makes future development easier
- adds some nice efficiency gains for joint inference of (x, theta) and deals with missingness there!

Special thanks to @theorashid for asking for many of these changes, then putting up an initial PR to make it happen!

theorashid:
Amazing job everyone. Great refactor

Le jeu. 30 juil. 2026 à 16:56, Matt Levine ***@***.***> a
écrit :

mattlevine22:
Happy factor day everyone!

PR #266, opened 2026-07-06
Simulator refactor

mattlevine22:
**Summary**

This PR finishes the simulator/latent-path refactor by separating forward generation from explicit latent-state inference.

The main architectural change is that simulators are now generation-only. Forward rollout lives in the simulator stack, while joint state/parameter inference now goes through `LatentPathBuilder`, backed by shared pure-JAX state-path reconstruction and scoring utilities.

Along the way, this also adds new standalone APIs and broadens the missingness/inference workflows supported by the library.

**Main changes**

- Simulators are now generation-only.
- `dsx.sample(..., obs_times=..., obs_values=...)` with simulator handlers no longer serves as a latent-state inference route.
- The old monolithic `dynestyx/simulators.py` implementation was split into a new `dynestyx/simulation/` package.
- Simulator execution is now pure JAX and returns `SimulatedResult`, with NumPyro site registration deferred to a `_register_numpyro_sites` step.
- Added standalone pure-JAX APIs in `dynestyx/api.py`:
  - `dsx.simulate(...)` for forward simulation
  - `dsx.log_prob(...)` for joint trajectory scoring
- Added `LatentPathBuilder` as the dedicated explicit latent-path inference handler.
- Added shared pure-JAX latent/state-path internals under:
  - `dynestyx/inference/latent/`
  - `dynestyx/inference/state_paths/`
- `LatentPathBuilder` is now NumPyro-facing only and supports `dsx.sample(...)`, not `dsx.condition(...)`.
- Simplified the latent-path execution flow so the builder now does a single reconstruction/scoring pass after dummy latent sites are created.
- Reworked missingness handling across latent-path inference:
  - partially missing `DiracIdentityObservation` data now automatically creates only the needed free state coordinates as latents
  - unsupported partially missing continuous observation models can fall back to explicit missing-observation augmentation through `f_missing_obs_values`
  - reconstructed completed observations are exposed through deterministic outputs such as `f_completed_obs_values`
  - ODE + Dirac inference is explicitly rejected
- Reorganized inference config modules under `dynestyx/inference/configs/`
- Moved inference helper utilities into `dynestyx/inference/utils/`
- Tightened the public API surface by removing top-level exports for low-level latent/missingness preparation helpers

**New capabilities**

- Pure-JAX forward simulation via `dsx.simulate(...)`, without needing a NumPyro handler stack
- Pure-JAX trajectory scoring via `dsx.log_prob(...)`, for evaluating `log p(x, y | ...)` on user-supplied latent paths
- A dedicated explicit latent-state inference route via `LatentPathBuilder`
- Deterministic ODE latent inference through `LatentPathBuilder`, where the latent path parameters are just the initial condition and the full path is reconstructed by solving the ODE
- Cleaner and broader missing-data support in latent-path inference:
  - exact-observation / Dirac partial missingness now infers only the missing state coordinates
  - unsupported correlated continuous observation families can use explicit observation augmentation
  - posterior samples of reconstructed missing observations are now directly available through `f_completed_obs_values`
- Cleaner composition of rollout with inference handlers, since simulators can now sit outside inference handlers as pure rollout/generation layers
- Unified simulator configuration objects (`SimulatorConfig`, `ODESimulatorConfig`, `SDESimulatorConfig`)
- Better support for custom pure-JAX workflows outside NumPyro, including manual optimization / inference loops

**A frustration (caching)**
One tricky design point in this refactor is latent-path layout preparation under NumPyro replay. LatentPathBuilder needs concrete observation missingness structure to determine latent dimensionality and indexing, but NumPyro’s MCMC/Predictive machinery may re-enter the model with traced values rather than ordinary arrays. That creates tension between three goals we want simultaneously: a simple fully hidden user API, compressed missingness-aware latent parameterizations, and replay-safe preparation under NumPyro tracing. The current cache-based solution is a pragmatic way to balance those goals: it preserves the simple API and compressed layouts without forcing users to precompute or pass bookkeeping objects themselves, at the cost of some extra internal complexity in this part of the implementation.

**Docs and tutorials**

- Updated simulator docs to reflect the generation-only contract
- Added dedicated public/developer docs for `LatentPathBuilder`
- Updated tutorials to use `LatentPathBuilder` instead of simulators for explicit latent inference
- Reworked `05_svi_no_numpyro` to use the new `dsx.simulate(...)` generation pattern
- Replaced the old simulator-based missingness MCMC tutorial with a latent-path version:
  - `11b_missing_observations_latent_path_mcmc.ipynb`
- Updated the missingness tutorial to show Dirac partial missingness and reconstructed observations under augmentation

**Tests**

- Added standalone coverage for:
  - `dsx.simulate(...)`
  - `dsx.log_prob(...)`
  - state-path assembly
  - `LatentPathBuilder`
  - simulator configs
- Updated simulator/inference/missingness smoke tests to the new API split


docs/api_reference/developer/inference/configs/filter_configs.md
DanWaxman, line 1: This should also be moved to `configs/`

docs/api_reference/developer/inference/configs/mcmc_configs.md
DanWaxman, line 1: Should be moved to `configs/`

docs/api_reference/developer/inference/configs/smoother_configs.md
DanWaxman, line 1: Should be moved to `configs/`

docs/api_reference/developer/handlers.md
```diff
@@ -4,4 +4,15 @@
...
+        - sample
+        - plate
+
+::: dynestyx.api
+    options:
+      show_root_heading: false
+      show_root_toc_entry: false
+      members:
+        - log_prob
+        - simulate
```
DanWaxman, line 18: Will have to think if this is the best place

docs/api_reference/developer/simulators.md
```diff
@@ -1,61 +1,76 @@
...
-`sample` sites for latent states and observations on a chosen time grid.
+Simulators (also called *unrollers*) run pure-JAX forward trajectories for a
+`DynamicalModel` on a chosen time grid, then optionally attach the realized
+outputs to a NumPyro trace when used through `dsx.sample(...)`.
+
+## Package Layout
+
+The simulator refactor split the previous monolithic `dynestyx/simulators.py`
+implementation into the `dynestyx/simulation/` package:
+
```
DanWaxman: No need for this here

DanWaxman (commented):
I changed the base branch from `main` to `dsx-infer-staging` and took a quick look. The organization looks reasonable.

1. There's a few places in docs where files should be moved around. 
2. It's surprising to me that `latent` has so much code now. I wonder if there's some boilerplate we can cut down on, or ways to unify the code. I get that missingness adds some complexity for us, but this amount of code is hard to digest and maintain.
3. I wonder the extent we need a type of callback machinery -- seems more complex than what goes on elsewhere.

docs/api_reference/developer/inference/latent_path_builder.md
```diff
@@ -0,0 +1,68 @@
+# LatentPathBuilder
+
+Developer-facing reference for latent-path construction and scoring.
+
+The latent-path refactor now separates responsibilities across two layers:
+
+- `dynestyx.inference.latent`: NumPyro-facing orchestration
+- `dynestyx.inference.state_paths`: pure-JAX state-path layout, reconstruction,
+  and scoring
```
DanWaxman:
```suggestion
The latent-path refactor separates responsibilities across two layers:

- `dynestyx.inference.latent`: NumPyro-facing orchestration
- `dynestyx.inference.state_paths`: pure-JAX state-path layout, reconstruction,
  and scoring
```

dynestyx/inference/configs/simulator.py
```diff
@@ -0,0 +1,105 @@
...
+class ODESimulatorConfig:
+    """Structured ODE simulator settings.
+
+    This config collects the solver settings used to reconstruct a deterministic
+    state path
+
+    ``x = (x(t_0), x(t_1), ..., x(t_T))``
+
+    from path parameters such as an initial condition. It is the preferred
+    replacement for passing raw Diffrax kwargs directly into ``ODESimulator``.
```
DanWaxman: don't mention replacement

dynestyx/inference/configs/simulator.py
```diff
@@ -0,0 +1,105 @@
...
+@dataclasses.dataclass
+class SDESimulatorConfig:
+    """Structured SDE simulator settings.
+
+    This config collects the backend and solver settings used to simulate an
+    SDE path. It supports either a Diffrax-based solve or the faster
+    Euler-Maruyama scan backend.
+    """
+
+    solver: dfx.AbstractSolver = dataclasses.field(default_factory=dfx.Heun)
```
DanWaxman, line 114: Maybe not for this PR, but it's a bit unfortunate that our default solver does not converge to the Ito solution.

dynestyx/inference/latent/builder.py
```diff
@@ -0,0 +1,430 @@
...
+coordinates three steps:
+
+1. prepare a concrete :class:`LatentPathLayout`,
+2. reconstruct and score ``x = g(z)`` in pure JAX, and
+3. optionally register NumPyro sites after that pure-JAX work is done.
+
+Keeping those responsibilities separate is what allows
+``LatentPathBuilder`` to keep its reconstruction/scoring logic in pure JAX
+even though the public handler itself is NumPyro-facing and only supports
+``dsx.sample(...)``.
```
DanWaxman: Maybe dumb question, but can things be simpler if they use numpyro from the beginning? I don't see a reason not to if `dsx.condition` is incompatible with the `LatentPathBuilder`...

DanWaxman (changes requested):
Some small comments here, but my overall bigger comment is that we seem to have many, many dataclasses here that propagate information locally, and probably aren't really necessary. This causes a lot of the bloat between definitions and instantiations. I'd also like to handle use under a `Predictive` better; currently, predictive will silently give you a bunch of Gaussians only for states. I have some ideas to fix this, but we should talk them through.

DanWaxman:
I opened #273 following some of our discussion on Friday regarding simplifying `LatentPathBuilder`. Much of this happens through inlining planners and introducing a new `ImproperUniform` subclass. Overview of that PR:

> This PR simplifies the LatentPathBuilder interface to a single `builder` file. It also fixes an issue with initializations.
> - Metadata is now resolved by an explicit caching mechanism
> - Lots of thin helpers were removed (e.g., for stacking plate results, which is a 3 line list comprehension)
> - The missingness mechanism is now through a sublcass of ImproperUniform, which simplifies some bookkeeping.
>   - This also fixes an issue with initializations for MCMC. Previously, `init_to_sample` would init to a fairly meaningless Gaussian sample. It now inits from a sample of the forward model of the dynamical system.
>   - Note that I'm not sure this is perfect... Realizations might be quite different, and one would often be better off with a lazy predictor initialization or something. Something to think about.
> - It also gets rid of canonicalize, since I don't really like that word. They're called `validation` methods now, which I also don't love, but dislike less.
> - Improves some documentation surrounding missingness metadata.
> - Adds some correspodning tests regarding new samples
> - Move a few utilities around, re-using simulation machinery where possible.
> 
> Some todos for remaining PR review include a closer look at observation missingness, a pass over simulation, and a better pass over some documentation and variable/method naming conventions.

mattlevine22:
We also discussed the idea of a new class that has .sample() implemented via a simulator and a .log_prob() implemented that does the stuff currently in LatentPathBuilder

Is the idea that this new PR would support that direction?

DanWaxman:
Yep!

dynestyx/inference/configs/simulator.py
```diff
@@ -0,0 +1,105 @@
...
+        tol_vbt_arr = (
+            dt0_arr / 2.0
+            if self.tol_vbt is None
+            else as_scalar_time_array(self.tol_vbt, name="tol_vbt")
+        )
+        if bool(tol_vbt_arr >= dt0_arr):
+            raise ValueError(
+                "tol_vbt must be smaller than dt0 for statistically correct simulation."
+            )
+        return tol_vbt_arr
```
DanWaxman, line 160: Is this used? Why does the `tol_vbt` not appear in the normalized `diffrax` settings?

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+            return leaf
+        if _leaf_is_plate_batched(leaf, plate_shapes, path=path):
+            return leaf[plate_idx]
+        return leaf
+
+    return jax.tree_util.tree_map_with_path(
+        _slice_leaf,
+        tree,
+        is_leaf=_is_opaque_plate_leaf,
+    )
```
DanWaxman: This feels like a lot of utilities for `base.py`

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+
+def _tile_times(times: Array, n_sim: int) -> Array:
+    """Return times tiled to shape (n_sim, T)."""
+    return jnp.broadcast_to(jnp.expand_dims(times, axis=0), (n_sim, len(times)))
+
+
+def _ensure_trailing_dim(arr: Array) -> Array:
+    """Ensure simulator outputs follow shape (n_sim, T, dim)."""
+    return arr[..., jnp.newaxis] if arr.ndim == 2 else arr
+
```
DanWaxman: Notably, some of the utilities aren't even used in `base.py`, e.g., this `_ensure_trailing_dim`

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+        "states",
+        "observations",
+        "predicted_times",
+        "predicted_states",
+        "predicted_observations",
+    ):
+        value = getattr(result, field_name)
+        if value is not None:
+            payload[field_name] = jnp.asarray(value)
+    return payload
```
DanWaxman: I don't really get what this is doing. It feels that it is just filtering out some `None`s and unused fields, but not sure why.

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+        "times": "times",
+        "initial_states": "x_0",
+        "states": "states",
+        "observations": "observations",
+        "predicted_times": "predicted_times",
+        "predicted_states": "predicted_states",
+        "predicted_observations": "predicted_observations",
+    }
+    for field_name, value in _simulated_result_site_payload(result).items():
+        numpyro.deterministic(f"{site_name}_{site_suffixes[field_name]}", value)
```
DanWaxman: Slightly unfortunate that these are different and need this extra layer of indirection, especially since we choose all the names lol

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+
+def _sample_initial_states(
+    initial_condition,
+    *,
+    rng_key: Array,
+    n_simulations: int,
+) -> Array:
+    """Draw independent initial states for each simulation member."""
+    keys = jr.split(rng_key, n_simulations)
+    return jax.vmap(initial_condition.sample)(keys)
```
DanWaxman: Rather thin wrapper

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+
+
+_SIMULATOR_CONFIG_UNSET = object()
```
DanWaxman: What does this do?

dynestyx/simulation/base.py
DanWaxman: nb. these are already moved in #273

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+
+    return jax.tree_util.tree_map_with_path(
+        _slice_leaf,
+        tree,
+        is_leaf=_is_opaque_plate_leaf,
+    )
+
+
+class BaseSimulator(ObjectInterpretation, HandlesSelf):
+    """Base class for generation-only simulator/unroller handlers.
```
DanWaxman:
```suggestion
    """Base class for generation-only simulator handlers.
```

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+            seg_names: list[str] = []
+            nonempty_seg_ids = [int(s) for s in np.unique(seg_ids_host)]
+            seg_keys = jr.split(rng_key, len(nonempty_seg_ids))
+            # Simulate one segment per present anchor (skip empty segments).
+            for seg_idx, seg_id in enumerate(nonempty_seg_ids):
+                # mask_host[i] = True iff predict_times[i] belongs to this segment id.
+                # This is the global-to-segment membership mask over the full prediction grid.
+                mask_host = seg_ids_host == seg_id
+                # Some segment ids may not own any prediction times. np.any here is host-side,
+                # avoids traced bool conversion, and lets us skip empty segment solves.
```
DanWaxman, line 196: Bit too much claude-speak for me

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+                                        initial_states=seg_result.initial_states
+                                    ),
+                                    site_name=seg_name,
+                                )
+                            )
+                        )
+                        for seg_name, seg_result in zip(
+                            seg_names, seg_results, strict=True
+                        )
+                    )
```
DanWaxman, line 248: I understand what's happening here, but would like to avoid this nested zipped comprehension stuff if possible. Feels like this construction could/should happen more directly.

dynestyx/simulation/base.py
```diff
@@ -0,0 +1,881 @@
...
+            if member_result is not None:
+                member_results.append(member_result)
+
+        if not member_results:
+            return None
+
+        return _stack_simulated_results(member_results, plate_shapes=plate_shapes)
+
+    @implements(_condition_intp)
+    def _sample_ds(
```
DanWaxman, line 405: `_condition_ds`?

dynestyx/simulation/ode.py
```diff
@@ -0,0 +1,176 @@
...
+import diffrax as dfx
+import jax
+import jax.numpy as jnp
+import jax.random as jr
+from jax import Array
+
+from dynestyx.inference.configs.simulator import ODESimulatorConfig
+from dynestyx.models import DynamicalModel
+from dynestyx.simulation.base import (
+    _SIMULATOR_CONFIG_UNSET,
```
DanWaxman: This pattern seems bad (it's a kinda weird way to have a semi-global variable)

dynestyx/simulation/ode.py
```diff
@@ -0,0 +1,176 @@
...
+                max_steps_value = 100_000
+            else:
+                max_steps_value = cast(int, max_steps)
+            simulator_config = ODESimulatorConfig(
+                solver=solver_value,
+                adjoint=adjoint_value,
+                stepsize_controller=stepsize_controller_value,
+                dt0=dt0_value,
+                max_steps=max_steps_value,
+            )
```
DanWaxman: Why can't we just construct an `ODESimulatorConfig` with the relevant fields and give `ODESimulatorConfig` the correct defaults?

dynestyx/simulation/sde.py
```diff
@@ -0,0 +1,273 @@
...
+
+class SDESimulator(BaseSimulator):
+    """Simulator for continuous-time stochastic dynamics (SDEs).
+
+    This simulator integrates a `ContinuousTimeStateEvolution` with nonzero diffusion
+    using Diffrax and a `VirtualBrownianTree` (see the Diffrax docs on
+    [Brownian controls](https://docs.kidger.site/diffrax/api/brownian/)). The
+    rollout itself is pure JAX; when used through `dsx.sample(...)`, realized
+    outputs are attached afterward as NumPyro deterministic sites, including
+    `"x_0"`, `"times"`, `"states"`, and `"observations"`.
```
DanWaxman: Don't we also have an `em_scan` version?

dynestyx/simulation/sde.py
```diff
@@ -0,0 +1,273 @@
...
+        controls are interpolated with a right-continuous rectilinear rule
+        (`left=False`), i.e., the control at time `t_k` is `ctrl_values[k]`.
+
+    Deterministic outputs:
+        When run through `dsx.sample(...)`, the simulator records `"x_0"`,
+        `"times"`, `"states"`, and `"observations"` as
+        `numpyro.deterministic(...)` sites.
+
+    Important:
+        - This is intended for **simulation / predictive checks** inside NumPyro.
```
DanWaxman: Now redundant, all `Simulator` objects are to be used this way

dynestyx/simulation/sde.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+            simulator_config = SDESimulatorConfig(
+                solver=solver_value,
+                stepsize_controller=stepsize_controller_value,
+                adjoint=adjoint_value,
+                dt0=dt0_value,
+                tol_vbt=tol_vbt_value,
+                max_steps=max_steps_value,
+                source=source_value,
+            )
```
DanWaxman: Same comment as ODE case about more direct default

tests/test_latent_path_builder.py
```diff
@@ -0,0 +1,895 @@
...
+    )
+
+
+def _make_deterministic_rollout_dynamics():
+    return dsx.DynamicalModel(
+        control_dim=0,
+        initial_condition=dist.Delta(jnp.array(0.0)),
+        state_evolution=lambda x, u, t_now, t_next: dist.Delta(x + 1.0),
+        observation_model=lambda x, u, t: dist.Delta(2.0 * x),
+    )
```
DanWaxman, line 107: I think we've more-or-less given up, but these should kinda-sorta be in `fixtures.py`

DanWaxman: Though again, I think it's worth it to just re-organize all our tests soon separately

tests/test_smoother_standalone.py
```diff
@@ -27,6 +29,17 @@ def _make_data():
...
+    return dsx.DynamicalModel(
+        control_dim=0,
+        initial_condition=dist.Normal(0.0, 1.0),
+        state_evolution=dsx.ContinuousTimeStateEvolution(
+            drift=lambda x, u, t: -0.1 * x
+        ),
+        observation_model=dsx.DiracIdentityObservation(),
+    )
+
+
```
DanWaxman, line 42: This at least should definitely be in `fixtures.py`, it's repeated in several test files

DanWaxman (changes requested):
Some more comments, mostly about `Simulator`s (so should be fairly independent of #273). This is generally very reasonable to me, with various style comments.

dynestyx/inference/configs/simulator.py
DanWaxman: Opened #274 , which resolve this as out of scope.

dynestyx/inference/configs/simulator.py
DanWaxman: This is because `diffeqsolve_settings` is actually for the `diffeqsolve` itself (the docomentation was a big misleading).

PR #269, opened 2026-07-10
Simplify state path planning

DanWaxman:
Previously, the state-path planner was quite complicated, consisting of three different sub-planning dataclasses. This commit consolidates them intoa single LatentPathLayout.


mattlevine22 (approved):
LGTM

PR #270, opened 2026-07-10
Add future forecasting functionality for LatentPathBuilder

mattlevine22:
Mimics how smoother creates distributions and forwards up to Simulator (only uses last time-point to avoid weird time/indexing shenanigans). Small refactor to avoid copy-pasting whole functions from smoothers.py.


DanWaxman (approved):
LGTM

PR #273, opened 2026-07-19
Simplify LatentPathBuilder interface

DanWaxman:
This commit simplifies the LatentPathBuilder interface to a single `builder` file. It also fixes an issue with initializations.
- Metadata is now resolved by an explicit caching mechanism
- Lots of thin helpers were removed (e.g., for stacking plate results, which is a 3 line list comprehension)
- The missingness mechanism is now through a sublcass of ImproperUniform, which simplifies some bookkeeping.
  - This also fixes an issue with initializations for MCMC. Previously, `init_to_sample` would init to a fairly meaningless Gaussian sample. It now inits from a sample of the forward model of the dynamical system.
  - Note that I'm not sure this is perfect... Realizations might be quite different, and one would often be better off with a lazy predictor initialization or something. Something to think about.
- It also gets rid of canonicalize, since I don't really like that word. They're called `validation` methods now, which I also don't love, but dislike less.
- Improves some documentation surrounding missingness metadata.
- Adds some correspodning tests regarding new samples
- Move a few utilities around, re-using simulation machinery where possible.

Some todos for remaining PR review include a closer look at observation missingness, a pass over simulation, and a better pass over some documentation and variable/method naming conventions.


dynestyx/inference/state_paths/reconstruct.py
```diff
@@ -114,36 +84,68 @@ def default_ode_diffeqsolve_settings() -> dict[str, Any]:
     }
 
 
-def assemble_completed_observation_state_path(
+def solve_ode_state_path(
```
mattlevine22: I like this, and I think it should replace most of our other uses of `_build_control_path` in simulators.py as well (for ODEs and SDEs).
- Let's put it in `solvers`, create an sde version, and do the replacements? 

Should reduce lines of code and improve organization

dynestyx/inference/latent/builder.py
```diff
@@ -179,160 +180,276 @@ def _sample_single(
...
+            raise ValueError(
+                "missing_observation_strategy must be one of 'auto', "
+                "'marginalize', 'augment', or 'error'."
+            )
 
-        1. prepare/canonicalize the latent request,
-        2. register the dummy latent NumPyro sites, and
-        3. reconstruct/score the path once from the active site values.
+        if obs_values_filled is None or obs_mask is None:
+            obs_values_filled, obs_mask, _ = prepare_observation_views(
```
mattlevine22: Didn't this already happen earlier in the stack?

dynestyx/inference/latent/builder.py
```diff
@@ -179,160 +180,276 @@ def _sample_single(
...
+            and (self.missing_observation_strategy in ("marginalize", "error"))
+        ):
+            raise ValueError(
+                "DiracIdentityObservation missingness in latent-path inference "
+                "supports only augment semantics. Use "
+                "missing_observation_strategy='auto' or 'augment'."
+            )
+
+        observation_dim = 1 if obs_values.ndim == 1 else obs_values.shape[-1]
+        use_observation_augmentation = False
```
mattlevine22, line 244: Shouldn't this be True? I thought observation augmentation is the cleaner way to do missingness in the exact-observation case

dynestyx/inference/latent/builder.py
```diff
@@ -179,160 +180,276 @@ def _sample_single(
...
         state_path_param_site = numpyro.sample(
             f"{name}_state_path_params",
-            state_path_param_base_dist,
-            obs=prepared.canonical_state_path_params,
+            _ForwardSimulationImproperUniform(
+                state_forward_sampler,
+                event_shape=state_event_shape,
+                sample_transform=state_sample_transform,
+            ),
+            obs=validated_state_path_params,
```
mattlevine22, line 321: So when `obs = (not None)`, this site just gets populated with `validated_state_path_params`? Sweet.

mattlevine22 (commented):
This is really great!

1. In `builders`, I want to check on the strategy when there is missingness + Dirac...I thought the better way to do this was to infer missing observations (then map the completed obs directly to the latent state).

2. There are some new solver functions like `solve_ode_state_path` that wrap a solver and add some control path eval stuff, etc. 
- I think we can organize these a bit more and use them throughout the codebase.

3. Then I think you wanted to clean up some more stuff with your other PR comments.
- should those go here? Or elsewhere?

dynestyx/inference/latent/builder.py
DanWaxman: Yes, happens in `condition` :) will remove

dynestyx/inference/latent/builder.py
DanWaxman: yessir

dynestyx/inference/latent/builder.py
DanWaxman: Now that we have semantically meaningful `sample` functions to the state prior, we have to do some work in constructing a transition distribution anyways. This new approach makes it so that state transition information is strictly performed _before_ observation augmentation, which I prefer.

dynestyx/inference/state_paths/reconstruct.py
DanWaxman: Done!

mattlevine22 (approved):
LGTM thanks!

PR #280, opened 2026-07-25
Update Documentation for Numpyro-Separated PR

DanWaxman:
This implements doc changes required by #255. 

- [x] Update API reference pages
- [x] Update other doc website pages (e.g., FAQ and mathematical introduction)
- [ ] Update notebook text
- [x] Rerun notebooks
- [x] Update docstring
- [x] Enforce `jaxtyping`


DanWaxman:
I'm doing a pass through docstrings today, @mattlevine22 feel free to do anything else but please avoid changing docstrings for now (or, better yet, enjoy your weekend 😆)

mattlevine22:
Sweet thanks @DanWaxman, I'll take a pass tomorrow on the train! Let me know if there are parts you don't get to, and I'll start there.

DanWaxman:
@mattlevine22 I just pushed [99c7051](https://github.com/BasisResearch/dynestyx/pull/280/commits/99c7051d504514ffffedb6d6434720149f4b2418); this handles docstrings for everything but the `simulation` directory. I have a checklist in the PR description that would be great to update status with.

It also includes two other small code changes (updating default arguments as per #281, and consolidating a helper in `observation_missingness` that seemed redundant to me). These seemed too small to me to be separate PRs.

mattlevine22:
Found a weird bug in our tests:

- All 28 tests in `tests/test_latent_path_builder.py` pass if run in serial, BUT `tests/test_latent_path_builder.py::test_latent_path_builder_bare_predictive_draws_dynamical_prior_paths` FAILS if run in isolation. 

Seems like this is related to pre-populating `_MISSING_OBSERVATION_METADATA_CACHE`.

@DanWaxman do you think this is a bug in our code or in our test (for depending on serial execution)?

DanWaxman:
I’d be inclined to say any test that passes in parallel but not serial is a
bug in the test. But tbf the global cache structure is a bit sketchy,
wouldn’t be surprised if that’s also an issue…

mattlevine22:
Re: passing in serial, yeah agreed. I'm going to leave it alone for now and come back to it.

Now, also noticing that `dsx.simulate` and `Simulators.simulate()` accept `obs_times` as an argument. Created #282 to track this and will do subsequent docstring updates there.

DanWaxman:
I ran all notebooks except `fhn_sparse_id.ipynb`, `gp_drift.ipynb`, and `lv_uode.ipynb`!

mattlevine22:
- I ran + pushed `fhn_sparse_id.ipynb` and  `lv_uode.ipynb` (gp_drift still running)
- I updated the docs, including api, FAQ, index, math intro.

I changed most of our docs to use `dsx.Simulator()` and `dsx.Filter()` instead of adding import lines AND instead of distinguishing sub-types of simulators...after all, no one should really need to use DiscreteTime/SDE/ODESimulator() directly anymore.

Will open a PR to deal with LatentPathBuilder dictionary argument nonsense.

dynestyx/inference/state_paths/reconstruct.py
```diff
@@ -27,11 +27,37 @@
 
 def validate_state_path_params(
     dynamics: DynamicalModel,
-    state_path_params: Array,
+    state_path_params: Real[Array, "state_path_param_time state_dim"]
+    | Real[Array, " _"]
```
mattlevine22, line 31: is this legit?

mattlevine22:
Asked codex to review the jaxtyping stuff, and it picked up on some things:

## 1. Plate-awareness contracts

These all stem from the same architectural mismatch: plate batching is handled by outer handlers, while the underlying pure computation helpers operate on one member at a time.

- **[P1] `dsx.simulate` falsely accepts plated arrays**  
  [dynestyx/api.py:33]
  `ctrl_times`, `ctrl_values`, and `predict_times` allow arbitrary leading plate axes, but `Simulator.simulate` requires unplated arrays.
---> We should probably restrict to non-plated and say "please use numpyro stack for that"?

- **[P1] `dsx.log_prob` falsely accepts plated observations and controls**  
  [dynestyx/api.py:102]
  The function passes its inputs directly to unplated reconstruction and scoring helpers.
---> We should probably restrict to non-plated and say "please use numpyro stack for that"?

- **[P2] `compute_hmm_filter` falsely advertises direct plate support**  
  [dynestyx/inference/hmm_filters.py:268]
  The implementation indexes and scans the leading axis as time. HMM plate batching happens outside this function in `Filter`, so its parameters and output should be unplated.
---> Here, I think we can just omit plate_shape from the jaxtyping.


## 2. Incorrect shape contracts

- **[P1] Batched Euler–Maruyama rejects scalar times**  
  [dynestyx/solvers/sde.py:171]
  The implementation explicitly broadcasts scalar `t_now` and `t_next`, but `Real[Array, "*batch"]` requires them to match the batch dimensions of `x`. Each should accept:
  ```python
  Real[Array, ""] | Real[Array, "*batch"]
  ```
  ---> Basically, it's saying that `t_now, t_next` don't have to also be batched!

- **[P2] Dense missing-observation values are rejected by the annotation**  
  [dynestyx/inference/state_paths/score.py:111]
  `prepare_observation_log_prob` supports dense completion arrays shaped `(time,)` and `(time, observation_dim)`, but this annotation only permits a flattened vector or scalar. The corresponding `dsx.log_prob` annotation should be widened too.
---> I'm a bit confused by this, but basically, I think it is saying that the code can also do dense completion arrays (not just what is missing)...I think we may use that for the more basic parts of the latent inference though. Not sure, a bit in the weeds here.

## 3. Incomplete jaxtyping migration

- **Euler–Maruyama return remains `dict[str, Array]`**  
  [dynestyx/solvers/sde.py:175]
  This loses the distinct `loc` and `cov` shapes. A typed dictionary would express them accurately.
---> eh, whatever?

- **HMM likelihood remains a raw `jax.Array`**  
  [dynestyx/inference/hmm_filters.py:318]
  It should be `Float[Array, ""]`.
  ---> yep.
  
- **`compute_hmm_filter` uses dtype-agnostic `Shaped` for the likelihood**  
  [dynestyx/inference/hmm_filters.py:278]
  This should likewise be `Float[Array, ""]`.
  ---> yep.

- **`Filter.marginal_loglik` remains a raw `jax.Array`**  
  [dynestyx/inference/filters.py:226]
  It should match `ConditionedResult.marginal_loglik`, currently `Real[Array, "*plate"]`.
  ---> yep.

mattlevine22 (approved):
Ship it off to Boston!

PR #281, opened 2026-07-25
Move to lax.map from custom chunked_vmap logic

DanWaxman:
The `_scan_chunked_vmap` is essentially a custom implementation of [`jax.lax.map`](https://docs.jax.dev/en/latest/_autosummary/jax.lax.map.html). This PR moves to `jax.lax.map` instead.


mattlevine22 (approved):
Looks great, tests pass, less code!

PR #282, opened 2026-07-26
Matt's docstrings and edits for no-numpyro

mattlevine22:
This PR:

1. removes `obs_times` stuff from the Simulator interface (`_sample_ds` in BaseSimulator still admits these arguments and gives a helpful error about this being deprecated).

2. Adds a long-form docstring to each of the Simulators (I followed the style of `Filter`'s docstring, which I thought was nice).
- I used AI with this, but was pretty prescriptive and suggested a lot of edits and read in detail.
- I'm quite happy with these docstrings
- At worst, they are too long (but honestly convey lots of useful information, albeit occasionally redundant across the docs, but I think this redundancy is in fact helpful...this is a simple one-stop-shop place to understand what you are doing when you run a Simulator() )
- I admittedly didn't quite check the correctness of the indexing of `"f_{j+1}_x_0"`, but I think it is right.

### DIFF
--- a/docs/api_reference/public/simulators/simulator_configs.md
+++ b/docs/api_reference/public/simulators/simulator_configs.md
@@ -1,0 +1,28 @@
+# Simulator Configurations
+
+Simulator configurations control how continuous-time trajectories are solved.
+Use `ODESimulatorConfig` for deterministic continuous-time dynamics and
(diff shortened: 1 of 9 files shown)


DanWaxman (approved):
Looks good to me -- we should probably also revamp the `api_reference/developer/simulation/*` pages, but I don't want that to block this finally landing 😆

PR #283, opened 2026-07-27
Bump version to 0.1.1

DanWaxman:
NumPyro-separation PR should be 0.2.0, since it's (very) breaking. Let's cut a 0.1.1 release, which is mostly documentation changes but also HMM missingness support, since the NumPyro-free version is nearing completion.


mattlevine22 (approved):
Sounds good

PR #284, opened 2026-07-27
LatentPathBuilder take ODESimulatorConfig as arg, not a dict

mattlevine22:


DanWaxman (changes requested):
Could you also update the ODE notebook, which defined these the old way?

mattlevine22:
The notebook doesn't specify settings at all, as far as I can see.

mattlevine22:
done, but accidentally pushed it directly to the parent branch. merging

mattlevine22 (commented):
Yep, although technically this commit happened BEFORE the merge w.r.t. dw-no-numpyro-docs (enabled running the ODE notebook).

PR #285, opened 2026-07-28
Remove caching

mattlevine22:
## Very AI PR ##

** Goal is to fix an issue where:
`tests/test_latent_path_builder.py::test_latent_path_builder_bare_predictive_draws_dynamical_prior_paths` fails, but running the whole test file in serial passes.

Previously, Dynestyx kept one shared list of “where observations are missing.” That list could accidentally come from another test or dataset with the same shape. This caused tests to pass or fail depending on what ran first.

Here we replace global missingness cache with handler-scoped metadata preparation

Prepare LatentPathBuilder missing-observation metadata automatically during
model tracing. For compiled MCMC replay, retain metadata only on the active
handler instance and validate the runtime mask before reuse.

This removes cross-model cache collisions and test-order dependence while
allowing standalone prior Predictive calls to work without cache warmup.

### DIFF
--- a/dynestyx/inference/latent/builder.py
+++ b/dynestyx/inference/latent/builder.py
@@ -2,15 +2,17 @@
 
 from __future__ import annotations
 
+import dataclasses
 import itertools
 
 import equinox as eqx
+import jax
 import jax.numpy as jnp
 import numpyro
 import numpyro.distributions as dist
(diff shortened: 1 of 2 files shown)


mattlevine22:
superseded by #286

PR #286, opened 2026-07-28
Some Bugfixes for NumPyro-Free Usage

DanWaxman:
Found some bugs from a Codex pass over #255.

- [Fix the filter configuration documentation link (was still linked to the old version).](https://github.com/BasisResearch/dynestyx/commit/7b5eac50a7fa20e12b5d74a3c7eabfedc3260c70)
- [Preserve smoother distributions during posterior rollout (posterior rollout previously deleted some information from metadata).](https://github.com/BasisResearch/dynestyx/commit/6b452823e1a018925adb58e144665fd6c0816b53)
- [Defer batched HMM site registration until sampling (HMMs retained some numpyro deterministic calls in condition).](https://github.com/BasisResearch/dynestyx/commit/1abb541735b6a1ad2433e8e56e7e3caf6be638da)
- [Clear cached filter and smoother state before each call (make sure attributed like log likeihood are reset each `dsx.sample` call with the same handler object).](https://github.com/BasisResearch/dynestyx/commit/72b7118ac9ca9348b92ba0f93f545799059e5e1a)
- [Broadcast compatible plate-member values before stacking (broadcast all plated objects before stacking them; necessary if there are different plating levels used).](https://github.com/BasisResearch/dynestyx/commit/9fdec59f252a19865d790cb5879775a98e6569ec)
- [Preserve distribution batch axes when summing path log probabilities (makes sure to preserve summing over time axis only).](https://github.com/BasisResearch/dynestyx/commit/0239323722c7a9fe47b3696ca2c44ae995422a33)
- [Keep Equinox runtime guards live under JIT (how we were using `eqx.error_if` didn't actually produce errors when jitted).](https://github.com/BasisResearch/dynestyx/commit/f86b98869eae614cfd0262c2f3b7a2b83f886609)
- [Remove global missingness metadata caching (use compile-time metadata resolution instead of caching mechanism).](https://github.com/BasisResearch/dynestyx/commit/d1a9b4d2e2fb9dc7695c02b4b36f1cfe3dd8682f)

### DIFF
--- a/docs/api_reference/developer/inference/latent_path_builder.md
+++ b/docs/api_reference/developer/inference/latent_path_builder.md
@@ -206,9 +206,18 @@ separate `missing_obs_values` site in this case. Exact missing observations
 support `"auto"` and `"augment"`.
 
 Missingness determines NumPyro site shapes, so its pattern must remain static
-while JAX traces a model. The builder evaluates closed-over observation values
-at compile time and derives missing-observation times from the current time
-array. Do not pass observation values as a dynamic argument to `jax.jit`.
+while JAX traces a model. The builder first attempts to infer the layout from
+concrete observations and caches it by sample site and observation shape. If
+the observations are traced, it uses that cache instead.
+
+For a cold outer-JIT call, pass metadata produced by
+`dsx.prepare_missing_observation_metadata(...)` through
+`dsx.sample(..., missing_obs_metadata=metadata)`. If neither inference nor the
+explicit or cached layout is available, the builder raises an error describing
+both remedies. Missing-observation times are always derived from the current
+`obs_times`; finite values and times may therefore change, while the dynamic
+mask is checked against the retained layout. One explicit metadata object is
+shared across plate members.
 
 ## Implementation modules
 
--- a/docs/api_reference/developer/inference/latent_path_builder.md
+++ b/docs/api_reference/developer/inference/latent_path_builder.md
@@ -205,26 +205,27 @@ to be inferred through `state_path_params`. The builder does not create a
 separate `missing_obs_values` site in this case. Exact missing observations
 support `"auto"` and `"augment"`.
 
-Missingness determines NumPyro site shapes, so its pattern must remain static
-while JAX traces a model. The builder first attempts to infer the layout from
-concrete observations and caches it by sample site and observation shape. If
-the observations are traced, it uses that cache instead.
+The missingness pattern can determine NumPyro site shapes, so it must remain
+static while JAX traces the model. With concrete observations, the builder
+caches the layout by sample site and observation shape.
 
-For a cold outer-JIT call, pass metadata produced by
-`dsx.prepare_missing_observation_metadata(...)` through
-`dsx.sample(..., missing_obs_metadata=metadata)`. If neither inference nor the
-explicit or cached layout is available, the builder raises an error describing
-both remedies. Missing-observation times are always derived from the current
-`obs_times`; finite values and times may therefore change, while the dynamic
-mask is checked against the retained layout. One explicit metadata object is
-shared across plate members.
+There are three supported paths: run concretely, reuse a cache from a concrete
+run, or pass `missing_obs_metadata` directly. The metadata can come from
+`dsx.prepare_missing_observation_metadata(...)` or a manually constructed
+`dsx.MissingObservationMetadata` object.
 
-When plate members have different missingness layouts, create the builder
-outside the model and reuse it so the per-member cache entries survive traced
-MCMC execution. Member-specific sample and deterministic sites retain their
-native shapes. If a ragged field cannot be stacked without changing those
-shapes, its aggregate `LatentStateResult` field is a flat list of per-member
-arrays, ordered with the rightmost plate index varying fastest.
+The recommended pattern defines the model first and applies
+`LatentPathBuilder` when the model runs. Cache reuse requires the same builder
+object for the concrete and traced calls. NumPyro MCMC routines typically make
+the concrete call automatically.
+
+The test suite covers all three paths and ragged MCMC. The builder derives
+missing-observation times from the current `obs_times`, so values and times can
+change while the layout stays fixed.
+
+For plate members with different layouts, use one builder for all traced calls.
+Ragged aggregate fields are flat lists of per-member arrays. Member sites keep
+their shapes, and the rightmost plate index varies fastest.
 
 ## Implementation modules
 
--- a/docs/api_reference/developer/inference/latent_path_builder.md
+++ b/docs/api_reference/developer/inference/latent_path_builder.md
@@ -219,6 +219,13 @@ both remedies. Missing-observation times are always derived from the current
 mask is checked against the retained layout. One explicit metadata object is
 shared across plate members.
 
+When plate members have different missingness layouts, create the builder
+outside the model and reuse it so the per-member cache entries survive traced
+MCMC execution. Member-specific sample and deterministic sites retain their
+native shapes. If a ragged field cannot be stacked without changing those
+shapes, its aggregate `LatentStateResult` field is a flat list of per-member
+arrays, ordered with the rightmost plate index varying fastest.
+
 ## Implementation modules
 
 - `dynestyx.inference.latent.builder` creates the sites, handles plates, records
--- a/docs/api_reference/developer/inference/latent_path_builder.md
+++ b/docs/api_reference/developer/inference/latent_path_builder.md
@@ -205,13 +205,10 @@ to be inferred through `state_path_params`. The builder does not create a
 separate `missing_obs_values` site in this case. Exact missing observations
 support `"auto"` and `"augment"`.
 
-Some JAX transformations trace a model before evaluating it. During tracing,
-the values in the observation mask may not be available to Python. The builder
-therefore creates `MissingObservationMetadata` during a normal model call and
-stores it for later traced calls. It matches the stored data by site name, its
-purpose, the observation shape, and the missing-observation strategy. If no
-match is available, run the model once with ordinary observation arrays before
-applying the JAX transformation.
+Missingness determines NumPyro site shapes, so its pattern must remain static
+while JAX traces a model. The builder evaluates closed-over observation values
+at compile time and derives missing-observation times from the current time
+array. Do not pass observation values as a dynamic argument to `jax.jit`.
 
 ## Implementation modules
 
--- a/dynestyx/inference/filters.py
+++ b/dynestyx/inference/filters.py
@@ -203,7 +203,7 @@ class Filter(BaseLogFactorAdder):
     There are several different filters available in `dynestyx`, each with their own strengths and weaknesses.
     What filters are applicable to a given model depends heavily on any special structure of the model (for example, linear and/or Gaussian observations).
     For a summary table of all config classes and when to use them, see
-    [Available filter configurations](filter_configs.md).
+    [Available filter configurations](configs/filter_configs.md).
 
     Defaults
     --------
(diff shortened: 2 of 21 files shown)


tests/test_smoother_standalone.py
```diff
@@ -110,8 +114,10 @@ def test_infer_smoother_does_not_register_numpyro_sites():
 def test_condition_smoother_no_observations():
     """dsx.condition with Smoother and no obs returns ConditionedResult with marginal_loglik=None."""
     dynamics = _make_lti_dynamics(0.5)
+    obs_times, obs_values = _make_data()
 
     with Smoother(smoother_config=KFSmootherConfig(filter_source="cuthbert")):
+        dsx.condition("observed", dynamics, obs_times=obs_times, obs_values=obs_values)
         result = dsx.condition(
```
mattlevine22, line 148: why twice?

tests/test_smoother_standalone.py
```diff
@@ -50,7 +50,11 @@ def test_infer_smoother_returns_infer_result():
 
     with Smoother(smoother_config=KFSmootherConfig(filter_source="cuthbert")):
         result = dsx.condition(
-            "f", dynamics, obs_times=obs_times, obs_values=obs_values
+            "f",
+            dynamics,
+            obs_times=obs_times,
+            obs_values=obs_values,
+            predict_times=jnp.arange(obs_times[-1], obs_times[-1] + 3.0),
```
mattlevine22, line 57: why include predict_times ?

tests/test_filter_standalone.py
```diff
@@ -127,8 +127,10 @@ def test_infer_does_not_register_numpyro_sites():
 def test_condition_no_observations():
     """dsx.condition with no obs returns ConditionedResult with marginal_loglik=None."""
     dynamics = _make_lti_dynamics(0.5)
+    obs_times, obs_values = _make_data()
 
     with Filter(filter_config=KFConfig(filter_source="cuthbert")):
+        dsx.condition("observed", dynamics, obs_times=obs_times, obs_values=obs_values)
         result = dsx.condition(
```
mattlevine22, line 161: why twice?

mattlevine22:
Codex report---do we want to update these error_if's too?

> 
The JIT guard fix is incomplete, though this issue already existed before the PR.

Unchecked return values remain at:

- `observation_missingness.py:370, 694`
- `inference/checkers.py:152, 161, 178, 185`

Because the returned checked arrays are discarded, JAX may remove these checks under JIT. The validator should return the checked value and the caller should use it afterward.

`inference/checkers.py:166` is a related warning case: traced execution currently skips the warning. The calls in `state_paths/score.py:174` and `state_paths/reconstruct.py:230` use static shape checks and are not affected.

mattlevine22:
So now we would raise with `jax.jit(predictive)(obs_values)`, right? Is that okay? just making sure explicitly

docs/api_reference/developer/inference/latent_path_builder.md
```diff
@@ -205,13 +205,10 @@ to be inferred through `state_path_params`. The builder does not create a
...
-the values in the observation mask may not be available to Python. The builder
-therefore creates `MissingObservationMetadata` during a normal model call and
-stores it for later traced calls. It matches the stored data by site name, its
-purpose, the observation shape, and the missing-observation strategy. If no
-match is available, run the model once with ordinary observation arrays before
-applying the JAX transformation.
+Missingness determines NumPyro site shapes, so its pattern must remain static
+while JAX traces a model. The builder evaluates closed-over observation values
+at compile time and derives missing-observation times from the current time
+array. Do not pass observation values as a dynamic argument to `jax.jit`.
```
mattlevine22: aye aye!

mattlevine22:
The following batch-stress-test fails---are we okay with that?

```python
def test_latent_path_builder_preserves_ragged_plate_metadata():
    dynamics = _make_dirac_discrete_dynamics()
    obs_times = jnp.array([0.0, 1.0])
    obs_values = jnp.array(
        [
            [[0.0, jnp.nan], [jnp.nan, 1.0]],  # two missing
            [[0.0, jnp.nan], [1.0, 2.0]],      # one missing
        ]
    )

    with trace(), seed(rng_seed=jr.PRNGKey(0)):
        with dsx.LatentPathBuilder():
            with dsx.plate("members", 2):
                result = dsx.sample(
                    "f",
                    dynamics,
                    obs_times=obs_times,
                    obs_values=obs_values,
                )

    assert jnp.array_equal(
        result.state_path_param_times[0],
        jnp.array([0.0, 1.0]),
    )
    assert jnp.array_equal(
        result.state_path_param_coordinate_indices[0],
        jnp.array([1, 0]),
    )
    assert jnp.array_equal(
        result.state_path_param_times[1],
        jnp.array([0.0]),
    )
    assert jnp.array_equal(
        result.state_path_param_coordinate_indices[1],
        jnp.array([1]),
    )
```

mattlevine22 (changes requested):
Found some still-hanging bugs. Let me know what you want to do about them, and we can go ahead with merge based on that.

No need to have perfect be enemy of good---but also, 1 day to squash some bugs is likely worth it, since we've been working hard to get this solid :D

tests/test_smoother_standalone.py
DanWaxman: This was because adding `predict_times` previously destroyed info in the `ConditionedResult`. Adding `predict_times` makes sure that doesn't happen.

tests/test_smoother_standalone.py
DanWaxman: This is because before, multiple `condition` calls under the same `Filter`/`Smoother` handler (including if you re-use the handler) would cache the log-likelihood. So this test would, without patches in this PR, have failed with the highlighted line added.

tests/test_filter_standalone.py
DanWaxman: This is because before, multiple `condition` calls under the same `Filter`/`Smoother` handler (including if you re-use the handler) would cache the log-likelihood. So this test would, without patches in this PR, have failed with the highlighted line added.

DanWaxman:
> So now we would raise with `jax.jit(predictive)(obs_values)`, right? Is that okay? just making sure explicitly

agh, that's pretty annoying, you're right. It's sad, I quite liked the `jax.ensure_compile_time_eval()` solution otherwise. will try to think of something that avoids this + the global cache, both of which are suboptimal to me.

tests/test_filter_standalone.py
mattlevine22: let's add a comment about this so we aren't left wondering later on

DanWaxman: done

DanWaxman:
I updated with reverting to cache but storing the cache at the `LatentPathBuilder` level + making it possible to specify with `prepare_missing_observation_metadata`.

> The following batch-stress-test fails---are we okay with that?

I think I'm happy with these being broadcast, what do you think? I agree that the outputs are not what I would fully expect, but the alternative was basically that the same test would crash before because it would just fail to return results. Given that these are just bookkeeping (right?) and don't affect the actual posterior computation, I'm inclined for this to be a future-us problem, as I think it entails changing a bit about how plated metadata is handled.

dynestyx/inference/filters.py
mattlevine22: these shouldn't have plates right?

mattlevine22: These shouldn't have plates right?

mattlevine22 (changes requested):
Some notes:

1. `ConditionedResult.states` is a bit inconsistent under `dsx.plate`; numpyro-side handles it fine.

```python

with dsx.plate(...):
    result = dsx.condition(...)

result.states
```

Unplated: states is the backend state object.
Plated HMM/cuthbert: states is (marginal_loglik, backend_state_object).
NumPyro inference and rollout are unaffected. It only hurts standalone consumers of plated result.states. Fine to mark nonblocking/follow-up, but it is still an inconsistent public return contract.

---> I think we are okay with this, right?

2. Outstanding Jax-typing in `filters.py`:
- `_filter_discrete_time` and `_filter_continuous_time` should not take plated dimensions.
- Should probably fix these and come back to more Jax-typing.
- This may stem from not having `integrations/` Jax-typed yet.
- also _smooth_discrete_time, _smooth_continuous_time

3. Codex thinks we should remove an "unused dense missingness handling scheme" in `observation_missingness.py` (and narrow `missing_obs_values` in `api.py: line 111`, `score.py:line 111` and `observation_missingness.py:line 722`
- I checked that indeed all tests still pass if I remove...

a) the block at [observation_missingness.py (line 778)]:

```python

if (
    scalar_observations
    and missing_obs_metadata is None
    and tuple(missing_obs_values.shape) == original_obs_shape
):
    missing_obs_values = missing_obs_values[:, None]
```

and b) the block in [observation_missingness.py (line 846)]

```python
if (
    metadata is None
    and missing_obs_values is not None
    and tuple(jnp.asarray(missing_obs_values).shape) == tuple(filled_obs.shape)
):
    completed_obs = jnp.where(obs_mask, filled_obs, missing_obs_values)
```

4. More on missingness across plates:

The following MCMC integration test fails when having a different number of missing observations across the 2 plate members. This seems worth fixing:

```python

def test_latent_path_builder_ragged_augmentation_mcmc_returns_member_sites():
    obs_times = jnp.array([0.0, 1.0])
    obs_values = jnp.array(
        [
            [
                [0.0, jnp.nan],
                [jnp.nan, 1.0],
            ],  # 2 missing
            [
                [0.0, jnp.nan],
                [jnp.nan, jnp.nan],
            ],  # 3 missing
        ]
    )

    def conditioned_model(obs_times=None, obs_values=None):
        alpha = numpyro.sample("alpha", dist.Uniform(0.0, 0.9))
        dynamics = _make_student_t_discrete_dynamics(alpha=alpha)

        with dsx.LatentPathBuilder(missing_observation_strategy="augment"):
            with dsx.plate("members", 2):
                dsx.sample(
                    "f",
                    dynamics,
                    obs_times=obs_times,
                    obs_values=obs_values,
                )

    mcmc = MCMC(
        NUTS(conditioned_model),
        num_warmup=2,
        num_samples=3,
        progress_bar=False,
    )
    mcmc.run(
        jr.PRNGKey(0),
        obs_times=obs_times,
        obs_values=obs_values,
    )

    samples = mcmc.get_samples()

    assert samples["alpha"].shape == (3,)
    assert samples["f_p0_missing_obs_values"].shape == (3, 2)
    assert samples["f_p1_missing_obs_values"].shape == (3, 3)
```

DanWaxman:
>     1. `ConditionedResult.states` is a bit inconsistent under `dsx.plate`; numpyro-side handles it fine.

I didn't really understand what this comment meant, other than HMM returns were a bit inconsistent; I fixed this. In general, I don't think we should care much about the "dsx.plate + numpyro-free" combination; `dsx.plate` intrinsically calls `numpyro.plate`.

>     2. Outstanding Jax-typing in `filters.py`:

These should be fixed, agree we should look again more carefully later. 

>     3. Codex thinks we should remove an "unused dense missingness handling scheme" in `observation_missingness.py` (and narrow `missing_obs_values` in `api.py: line 111`, `score.py:line 111` and `observation_missingness.py:line 722`

I think this is a test coverage issue instead of a dead code issue? Would prefer to leave for now, personally.

>     4. More on missingness across plates:
> 
> 
> The following MCMC integration test fails when having a different number of missing observations across the 2 plate members. This seems worth fixing:

This is because of the stacking thing still. I changed it so that ragged quantities are returned as a flat list if they are not unstackable. This also gets rid of the wacky broadcasting semantics that were causing issues before.

mattlevine22 (approved):
Rock and roll!!!

PR #291, opened 2026-07-30
Control v2

### LINKED ISSUE #288: Discrete time, closed loop control
This issue aims to implement support for discrete time, closed loop control in Dynestyx.

# Control Loop
Basic control loop.

$$
\begin{aligned}
x_0 &\sim p(x_0 ; \rho), \\
y_0 \mid x_0 &\sim p(y_0 \mid x_0, t_0 ; \rho), \\
\hat{x}_{0 \mid 0} &= \mathrm{FilterUpdate}(y_0, t_0 ; \eta), \\
u_k, s_{k+1} &= \pi(\hat{x}_{k \mid k}, s_{k} ; \gamma), \qquad k = 0, \dots, T-1, \\
x_{k+1} \mid x_k &\sim p(x_{k+1} \mid x_k, u_k, t_k, t_{k+1} ; \rho), \qquad k = 0, \dots, T-1, \\
y_{k+1} \mid x_{k+1} &\sim p(y_{k+1} \mid x_{k+1}, u_k, t_{k+1} ; \rho), \qquad k = 0, \dots, T-1, \\
\hat{x}_{k+1 \mid k+1} &= \mathrm{FilterUpdate}(\hat{x}_{k \mid k}, u_k, y_{k+1}, t_k, t_{k+1} ; \rho, \eta), \qquad k = 0, \dots, T-1.
\end{aligned}
$$

The implementation must satisfy the following constraints:
1. Discrete time only
2. The transition probability $\rho$ should be any black-box callable (e.g. MuJoCo simulation, ODE/SDE, a dynestyx model...)
3. The policy $\pi$ should be a callable (e.g. equinox Module with callable).

Should return:
1. Filtered states $x_{0:N}$.
2. Observations $y_{0:N}$.
3. Policy states $s_{0:N}$ (useful for diagnostics). 
4. Control sequence $u_{0:N}$.

**Current limitations**
`DiscreteTimeSimulator` requires `ctrl_values` ahead of time (open loop controller).
`FilterUpdate` requires a full trajectory.

Potential issues:
1. Compatibility with `jax.lax.scan`, `jit`, `jax.grad`: true blackbox $\rho$ will break this.

# Basic control algorithms

This issue also aims to add support for basic controls in dynestyx (namely MPPI or other sampling based methods).  

#291

MatthieuDarcy:
This PR adds at first attempt at support for closed loop control in dynastyx. 

Added in /dynestyx/control 

1. `DiscreteControlLoopSimulator`

API:

```
    def model():
        with DiscreteControlLoopSimulator(
            control_policy=policy,
            policy_state_init=None,
            filter_config=FilterConfig(record_filtered_states_mean=True),
        ):
            return dsx.sample("loop", dynamics, predict_times=predict_times)
```

2. `MPPI`

Basic MPPI controller

API
```
mppi = MPPI(
    dynamics_model=dynamic_rollout
    loss_fn=quadratic_loss,
    horizon=horizon,
    n_samples=n_samples,
    noise_std=1.0,
    temperature=1.0,
)
```

Added in tutorials 
3. Notebook controller_demo.ipynb

Additional elements I will work on

1. Verifying compatibility with different filters.
2. Verifying gradient computation.

`Closes #288 `


MatthieuDarcy:
The biggest addition to other parts of the code is in `cuthbert/discrete_filter.py`

`compute_cuthbert_filter_update` now computes a one step filtering update: takes in the previous estimated state, a control, and observation and provides a new estimated state.

mattlevine22 (commented):
This looks great, thank you @MatthieuDarcy !

1. Can we have @cooijmanstim look at the Jupyter notebook demo, and get his take?

2. My preliminary comments/questions (haven't looked carefully through the code yet):

- I think `DiscreteControlLoopSimulator` should be made available directly through the `Simulator()` handler via `simulation/auto.py` (and, hence, numpyro-free `dsx.simulate`) interface.
- Should check that we can autodiff through a ClosedLoopSimulation.
- Should probably support `n_simulations > 1`.
-`DiscreteControlLoopSimulator` should another argument for an "approximate model" (assuming the "true" model is in the `dsx.sample` statement)?
- I like that one can feed very generic controllers (including MPPI); however, I think the current MPPI setup should be better curated so that the user doesn't have to do a bunch of the extra manual work that appears in the demo. I think the following should be possible:

```python

sim_mppi = DiscreteControlLoopSimulator(
    control_policy_config=MPPIConfig(
                                                      loss_fn="quadratic",  # or a callable
                                                      horizon=horizon,
                                                      n_samples=n_samples,
                                                      noise_std=1.0, # is this needed?
                                                      temperature=1.0,
                                                ),
    policy_state_init=mppi_initial_state(horizon, control_dim),
    filter_config=KFConfig(record_filtered_states_mean=True),
)

# MPPI should have make_mppi_rollout internally (don't expose user to it)
```

cooijmanstim:
Looks great, this interface works for us. For the notebook it would be good besides no control vs control to see the effectiveness of the controller as the noise level on the observations changes. Just three variants in total will do (no control, control with lots of noise, control with little noise).

docs/tutorials/control/controller_demo.ipynb
DanWaxman, line 1: Some notes*:

- $x_hat_{k|k}$ in the control policy should be $\hat{x}_{k|k}$.
- "with $a=1.05$, is linear with additive Gaussian noise. Without control, this system is unstable and $x_k \rightarrow \infty$" depends on the initial condition (if $x_0 < 0$, then $x_k \to -\infty$).
- We should probably package a few things, e.g., linear policy, so that the user does not have to recreate it in the simplest cases
- I think I would prefer if we follow the more general `dynestyx` pattern of using `with DiscreteControlLoopSimulator(): dsx.sample(...)`, though what you wrote also works
- What does $x_t^2$ mean in the SDE example (I think $x_t \odot x_t$?)
- "Both _cuthbert_filter_pf and _cuthbert_filter_enkf only call dynamics.state_evolution(...).sample(key)" this is more internal than I'd want to include in a tutorial
- MPPI is enough of a jump that it might be worth making a genuinely follow-up notebook

\* I understand a lot of this was probably preliminary text anyways, just figured it might be useful since a few of them are more general patterns

docs/tutorials/control/control_optimization.ipynb
DanWaxman, line 1: Would it make more sense to do LQR here?

dynestyx/control/discrete_controller_simulators.py
DanWaxman, line 1: I think it might be beneficial to just implement these directly in `Simulator` classes. The issue here is just that we made a split to actively recommend against `Simulator` for parameter inference...

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -151,6 +152,134 @@ def _drop_if_time_leaf(leaf):
...
+    only valid when the whole control trajectory is already known in advance.
+    For online control this is impossible: u_{k+1} cannot exist before
+    y_{k+1} is observed, since it is computed by the policy from the filtered
+    state that itself depends on y_{k+1}. So `u` here is used for both
+    `CuthbertInputs.u` and `CuthbertInputs.u_prev` in the single-row input
+    built for this step. Pass `u=None` for the bootstrap call, matching y_0's
+    lack of a control argument in the control-loop equations (numerically
+    equivalent to zeros for models with a control-input matrix, since D=None
+    or u=None are both treated as "no control contribution").
+    """
```
DanWaxman: too much docstring for an internal function

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -151,6 +152,134 @@ def _drop_if_time_leaf(leaf):
...
+    dynamics: DynamicalModel,
+    filter_config: BaseFilterConfig,
+    prev_state,
+    key: jax.Array,
+    *,
+    y: jax.Array,
+    u: jax.Array | None,
+    t: jax.Array,
+    t_prev: jax.Array | None = None,
+):
```
DanWaxman: can probably change this a bit and then scan over it instead of duplicating prepare/combine code?

DanWaxman (dismissed):
This looks like a good start! Some early high-level comments.

docs/tutorials/control/control_optimization.ipynb
MatthieuDarcy: I will take a look at some implementation of LQR, but I actually think MPPI is the better first example (typically more robust and works better on a broader range of problems for my understanding).

DanWaxman: Sounds good! I think if the example is a linear policy on a linear SSM, might as well do the LQR loss function, but doing MPPI as the main tutorial is also fine by me.

MatthieuDarcy:
Addressed some of the points made (thanks @mattlevine22 and @DanWaxman !). 
# Changes 
1. `DiscreteControlLoopSimulator` now works with `dsx.simulate`
Example syntax:

``
dsx.simulate(
        dynamics,
        rng_key=key,
        predict_times=predict_times,
        control_policy=policy,
        filter_config=KFConfig(record_filtered_states_mean=True)
    )
``

2. <strike> Policy is any callable that takes in `x_hat` (estimated filtered state) and `s` (internal state) and returns `u` and `s_new`.

`u` can be an array or a numpyro distribution, in which case the simulator will draw a sample from the distribution.
</strike>
3. MPPI is more robust and has better defaults. It's only required parameters at initialization are `loss_function` and `dynamics` (thus allowing for dynamics different from the dynamics evolving the state itself).

4. The notebooks `controller_demo.ipynb` and `mpc_demo.ipynb`have been improved and now expose some of these choices. The MPPI example is now exclusively in `mpc_demo.ipynb`. 

Explained in the notebooks are: 

- how to define a policy
- how to use `DiscreteControlLoopSimulator` with non-linear dynamics
- how to use ``MPPI` (highlighting that the dynamics used by MPPI have to be defined outside the simulator and thus can be different from the ones used by the simulator).

# Outstanding questions

Currently policy does not use time $t$.  This actually might be relevant (such as MPC combined with time dependent dynamics). Should we have policies take 3 arguments instead?

The simulator needs an initial state $s_0$. Internally, the simulator will check if `policy` has an `initial_state` attribute, defaulting to `None` otherwise. This makes it more convenient for the user to implement simple policies. In the notebook, I have the policy explicitly return `None` to highlight that, in general, policies might need to return an initial state. Is this alright?

# Todos

1. Address some of @DanWaxman's requested changes, especially improving the internal docs.
2. <strike>Simplify the test suite.</strike>
3. <strike>Test gradients and policy optimization.</strike>
4. Expand the policies we package.

MatthieuDarcy:
Update:

1. The control loop requires that the policy be any callable (e.g. a learned neural policy, an model predictive controller...) that accepts 4 arguments: 

* `x_hat` <strike>(an estimate of the filtered state, provided by the filter) </strike> (a numpyro distribution infered from the states).
* `t_now` the current time.
* `t_next` the next time at which the dynamics will advance.
* `s` internal state.

It must return 
* A control `u`, a `jax.numpy.array`.
* A new state `s`.

If `u` is a distribution, will raise an error and instruct the user to sample from the distribution instead. Tutorial have been updated.

# Sharp edges

The state returned by the filter must be convertible to a numpyro distribution. The current implementation uses `filter_state_dist` to do this: supports (mean, cov) -> Gaussian and particles -> weighted/uniform particles. Other filters might break this.

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: Cleaned this up. Now can either receive a config or a filter object (but not both, will raise an error). The ControlDiscretizer builds the filter outside the loop/scan and passes it to `compute_cuthbert_filter_udpate`.

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: Cleaned this up.

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Done

docs/tutorials/control/controller_demo.ipynb
MatthieuDarcy: These should all be fixed (except for 3, I will defer this to when we have converged to a structure we like).

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -151,6 +152,142 @@ def _drop_if_time_leaf(leaf):
...
+    )
+    if parallel and not filter_obj.associative:
+        raise ValueError(
+            "Associative filtering was requested, but the constructed cuthbert "
+            f"filter is not associative: {type(filter_config).__name__}."
+        )
+    return filter_obj, parallel
+
+
+def build_cuthbert_filter(
```
mattlevine22: this is a classic "thin function" that LLMs like to build. I think it is only twice and is basically 2 lines of code!

Moreover, maybe its usage can be removed from `compute_cuthbert_filter_update` (why not always pass that a filter object?)

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -151,6 +152,142 @@ def _drop_if_time_leaf(leaf):
...
+    filter_kwargs = _config_to_filter_kwargs(filter_config)
+    filter_obj, _ = _build_cuthbert_filter_obj(
+        dynamics, filter_config, filter_kwargs, key, want_parallel=False
+    )
+    return filter_obj
+
+
+def compute_cuthbert_filter_update(
+    dynamics: DynamicalModel,
+    filter_config: BaseFilterConfig | None,
```
mattlevine22: Why have both options? Should probably just take a filter object right? @DanWaxman

dynestyx/inference/integrations/cuthbert/discrete_filter.py
mattlevine22: Seems best to pick one contract---and a filter object seems like a good one. Any reason to keep the config option here?

dynestyx/control/discrete_controller_simulators.py
mattlevine22: > I think it might be beneficial to just implement these directly in Simulator classes. The issue here is just that we made a split to actively recommend against Simulator for parameter inference...

This is okay because we aren't recommending/supporting numpyro-based parameter inference; just auto differentiable simulation results.

mattlevine22 (changes requested):

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: I agree, I would require to pass the filter itself.

dynestyx/inference/integrations/cuthbert/discrete_filter.py
mattlevine22: put differently, I would scrap `_build_cuthbert_filter_obj` and put it all inside this function (and let this function also return parallel)

MatthieuDarcy: "I would scrap _build_cuthbert_filter_obj and put it all inside _this function_." Could you clarify which function you're referring to? 

I personally like the thin function, I think it makes other parts of the code more readable. Happy to remove it otherwise.

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -151,6 +152,142 @@ def _drop_if_time_leaf(leaf):
...
+        want_parallel
+        and isinstance(filter_config, KFConfig)
+        and filter_config.associative
+    )
+    if parallel and not filter_obj.associative:
+        raise ValueError(
+            "Associative filtering was requested, but the constructed cuthbert "
+            f"filter is not associative: {type(filter_config).__name__}."
+        )
+    return filter_obj, parallel
```
mattlevine22, line 228: I would just cut all of this code and drop it into `build_cuthbert_filter`

dynestyx/inference/integrations/cuthbert/discrete_filter.py
mattlevine22: described above. I'd keep this function and get rid of `_build_cuthbert_filter_obj`

MatthieuDarcy: Got it, will remove `_build_cuthbert_filter_obj` (this will also require to update `compute_cuthbert_filter`).

MatthieuDarcy:
@mattlevine22 @DanWaxman what do you think of the PR in it's current state regarding the DiscreteControlLoopSimulator only?

I have updated MPPI so that the loss function takes in a `ControlledSimulationResult` and returns a `scalar`. The implementation is clunky under the hood but works. My feeling is that it would be nice to have the overall structure merged and to create a seperate PR to impore the MPPI implementation (as this is separate and might involve changes to other parts of the code) and other MPC algorithms. Thoughts?

mattlevine22:
1. I opened PR #313 to try to update some stuff here.
- main things were to beef up the typing, add API documentation, do a few more checks / error-raises, and massage some of the variable names to be more clear/consistent
2. Realized that there is an off-by-one incompatibility of the ClosedLoopSimulator vs `dsx.simulate` when called with the selected controls specifically when the observation depends on a control. I opened issue #312 about this. I'm thinking we let this slide for now, and fix it more generally separately.
3. I still don't love that this has a `TYPE_CHECKING` variable to avoid circular imports, but oh well haha.

mattlevine22 (approved):
@DanWaxman is happy and so am I ! Let's do this thing---congrats @MatthieuDarcy on your first big Dynestyx / Basis PR haha, great work! Let's hope @cooijmanstim and the R-ADA team enjoy the fruits of your labor :) I'm sure many controls people will !

PR #293, opened 2026-07-31
Adding jaxtyping and small shape improvements to inference

mattlevine22:
- Add jaxtyping across Filter/Smoother inference.
- Canonicalize scalar observation/control series from (T,) to (T, 1).
- Preserve (T,) scalar labels for HMMs.

### DIFF
--- a/dynestyx/inference/filters.py
+++ b/dynestyx/inference/filters.py
@@ -14,7 +14,6 @@ from jaxtyping import Array, Bool, PRNGKeyArray, Real
 
 from dynestyx.handlers import HandlesSelf, _condition_intp
 from dynestyx.inference.checkers import (
-    _ensure_trailing_event_axis,
     _validate_batched_plate_alignment,
     _validate_inference_supported_model_classes,
     _validate_missing_observation_support,
@@ -75,7 +74,7 @@ from dynestyx.types import (
     FunctionOfTime,
     chain_numpyro_site_registrations,
 )
-from dynestyx.utils import _dist_has_plate_batch_dims
+from dynestyx.utils import _dist_has_plate_batch_dims, _ensure_trailing_event_axis
 
 type SSMType = ContDiscreteNonlinearGaussianSSM | ContDiscreteNonlinearSSM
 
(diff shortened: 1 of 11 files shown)


dynestyx/inference/filters.py
```diff
@@ -275,6 +276,16 @@ def _add_log_factors(
...
+        if not isinstance(config, HMMConfigs):
+            obs_values = _ensure_trailing_event_axis(
+                obs_values,
+                plate_shapes=plate_shapes,
+            )
+            if ctrl_values is not None:
+                ctrl_values = _ensure_trailing_event_axis(
+                    ctrl_values,
+                    plate_shapes=plate_shapes,
+                )
```
DanWaxman: Codex points out (and I agree) that this can be problematic to do *before* plating; for plate size `M` and time series length `T`, if `M = T`, then shared scalar controls `(T,)` get promoted to plated controls `(T, 1)`, which gets interpreted as plated. We should either do this later, or raise an error in `_ensure_trailing_event_axis` for this (rather rare) edge case.

DanWaxman (changes requested):

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
```diff
@@ -136,13 +139,13 @@ def _run_linear_kf(
...
-    obs_times: jax.Array,
-    obs_values: jax.Array,
-    ctrl_times=None,
-    ctrl_values=None,
-):
+    obs_times: Real[Array, " obs_time"],
+    obs_values: Real[Array, "obs_time observation_dim"],
+    ctrl_times: Real[Array, " ctrl_time"] | None = None,
+    ctrl_values: Real[Array, "ctrl_time control_dim"] | None = None,
+) -> Any:
```
DanWaxman, line 148: I'd like to avoid `Any` as return types. This need not be blocking for now, though (I think it's a bit annoying because `cd_dynamax` types are a bit all over the place).

DanWaxman (approved):
I made a few tiny nitpicks and pushed. LGTM otherwise, please check my changes then feel free to merge!

dynestyx/inference/integrations/cd_dynamax/continuous_filter.py
mattlevine22: agreed, but yeah let's table it. Thanks!

PR #298, opened 2026-08-07
Support for sparse observation matrices

MatthieuDarcy:
This PR adds support for sparse observation matrices H from https://docs.jax.dev/en/latest/jax.experimental.sparse.html.

Fairly small changes. 

- PF works as is. 
- EnKF: simply replaced `jnp.dot` with `@` and added a check before `jnp.asarray`.
- EKF: compatible out of the box, but likely no efficiency gains due to autograd. Will raise a warning.
- KF: incompatible due to Cuthbert linalg implementation (unclear if this can be cleanly fixed). Will raise an error.

Added 3 tests to check that computations give the same answers and that warnings/errors are correctly raised.

### DIFF
--- a/dynestyx/inference/filters.py
+++ b/dynestyx/inference/filters.py
@@ -11,7 +11,6 @@ import numpyro
 from cd_dynamax import ContDiscreteNonlinearGaussianSSM, ContDiscreteNonlinearSSM
 from effectful.ops.semantics import fwd
 from effectful.ops.syntax import ObjectInterpretation, implements
-from jax.experimental import sparse as jax_sparse
 from jaxtyping import Array, Bool, PRNGKeyArray, Real
 
 from dynestyx.handlers import HandlesSelf, _condition_intp
@@ -71,7 +70,6 @@ from dynestyx.inference.utils.plate_utils import (
     _slice_dist_for_plate_member,
 )
 from dynestyx.models import DynamicalModel
-from dynestyx.models.observations import LinearGaussianObservation
 from dynestyx.types import (
     ConditionedResult,
     FunctionOfTime,
@@ -279,25 +277,6 @@ class Filter(BaseLogFactorAdder):
                 mode="filter",
             )
 
-        obs_model = dynamics.observation_model
-        # Sparse observation matrices: incompatible with KF, compatible with EKF (but most likely inefficient)
-        if isinstance(obs_model, LinearGaussianObservation) and isinstance(
-            obs_model.H, jax_sparse.JAXSparse
-        ):
-            if isinstance(config, KFConfig):
-                raise ValueError(
-                    "A sparse observation matrix H was passed to KFConfig. This is not "
-                    "supported: the underlying cuthbert Kalman-filter internals cannot "
-                    "handle a sparse H. Pass a dense H, or use EnKFConfig/EKFConfig instead."
-                )
-            if isinstance(config, EKFConfig):
-                warnings.warn(
-                    "A sparse observation matrix H was passed to EKFConfig. This works "
-                    "correctly, but likely gives no efficiency gain as EKF differentiates the observation log-density instead, which"
-                    "scales with the dense state/observation dimensions "
-                    "regardless of H's sparsity.",
-                    stacklevel=2,
-                )
         # Resolve PRNG key: use explicit seed from config, fall back to numpyro
         # context (inside a seeded model), or None (deterministic filters don't need one).
         if config.crn_seed is not None:
--- a/dynestyx/inference/filters.py
+++ b/dynestyx/inference/filters.py
@@ -1,5 +1,6 @@
 import dataclasses
 import math
+import warnings
 from abc import ABC, abstractmethod
 from typing import cast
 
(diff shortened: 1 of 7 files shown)


dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -405,7 +406,13 @@ def get_observations(mi: CuthbertInputs):
...
         if isinstance(obs_model, LinearGaussianObservation):
             obs_params = obs_model.params_at(mi.time)
-            H = jnp.asarray(obs_params.H)
+
+            # Handle sparse observation matrix
+            if isinstance(obs_params.H, jax_sparse.JAXSparse):
+                H = obs_params.H
+            else:
+                H = jnp.asarray(obs_params.H)
+
```
DanWaxman, line 423: I think we can just be less defensive and use `H = obs_params.H`?

dynestyx/inference/filters.py
```diff
@@ -276,13 +279,30 @@ def _add_log_factors(
...
+        # Sparse observation matrices: incompatible with KF, compatible with EKF (but most likely inefficient)
+        if isinstance(obs_model, LinearGaussianObservation) and isinstance(
+            obs_model.H, jax_sparse.JAXSparse
+        ):
+            if isinstance(config, KFConfig):
+                raise ValueError(
+                    "A sparse observation matrix H was passed to KFConfig. This is not "
+                    "supported: the underlying cuthbert Kalman-filter internals cannot "
+                    "handle a sparse H. Pass a dense H, or use EnKFConfig/EKFConfig instead."
+                )
```
DanWaxman: Will KF with `backend="dynamax"` work?

DanWaxman (changes requested):
Looks awesome! Few small comments (and also the failing lint/tests, which I suspect may be partially because of a new `cuthbert` release; will look soon).

dynestyx/inference/filters.py
MatthieuDarcy: Will check

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: What was the logic of the original `H = jnp.asarray(obs_params.H)`? Is it just a LLM habit?

DanWaxman: Yeah mostly it's the LLMs being very defensive, I think.

dynestyx/inference/filters.py
MatthieuDarcy: Seems it does @DanWaxman. We have two options:

1.  raise an error and ask that the user specify in the filter config that filter_source should be dynamax 
2. raise a warning and state to the user that we are calling dynamax
3. just do it silently

I think 2 is the more elegant and user friendly option

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: ok, will remove.

dynestyx/inference/filters.py
DanWaxman: I'd prefer 1 personally (especially because there are options -- namely, for associative filters -- that exist for cuthbert but not dynamax). @mattlevine22 do you have thoughts?

mattlevine22: I'm more in line with 1 (require dynamax for sparse KF filtering).

Also, this I'd prefer this chunk of logic not live inside the already long `Filter` class.

Let's either move it to:
a) a chunk of logic in `_validate_sparse_filter_support(dynamics, filter_config)` 
b) deeper into integrations/cuthbert/discrete_filter.py (KF error, EKF warning) + integrations/cd_dynamax/discrete_filter.py (EKF warning)

- so far, we've been doing library-level checks inside the integrations code, so while (a) seems cleaner, (b) may follow our existing convention. I'm fine with either, though.

mattlevine22 (changes requested):
Overall looks good, added a comment about how to check/error on filter configs and where to put the code for it

dynestyx/inference/filters.py
MatthieuDarcy: For a) I can add a checker in inference/checkers and call it inside`_add_log_factors`. One thing is that currently we don't do anything for smoothers.

I think I actually prefer b) because it should handle both the filter and the smoother. I also need to run some additional tests to verify all backend-filter combinations.

mattlevine22: Let's go for (b) then, it is my preference as well.

MatthieuDarcy: Done. Can confirm that cd_dynamax supports all versions (EKF has the same issue regardless: unlikely to provide computational gains).

DanWaxman:
One other thing here: sparse matrices break lots of our plating logic. For now, I am fine personally with detecting sparse matrices in plating validation and raising an error + opening an issue, I don't want to hold this PR hostage if the main idea is its benefit for PDEs.

MatthieuDarcy:
> One other thing here: sparse matrices break lots of our plating logic. For now, I am fine personally with detecting sparse matrices in plating validation and raising an error + opening an issue, I don't want to hold this PR hostage if the main idea is its benefit for PDEs.

It's not critical, but some observation operators are very naturally sparse matrices so it would be nice to have.

DanWaxman (approved):
I added a guard for plating in [05a0998](https://github.com/BasisResearch/dynestyx/pull/298/commits/05a09984d56fd6f0127423cd242a88eac6719167) and opened #308. This is good on my end now. Thanks Matthieu!!

DanWaxman:
@mattlevine22 do you have any remaining concerns? I'm satisfied to merge this in main, personally.

mattlevine22 (approved):
Looks great thanks for this update @MatthieuDarcy and for the plating-check-patch @DanWaxman !

PR #300, opened 2026-08-07
Update to cuthbert v0.0.14

### LINKED ISSUE #302: Upgrade cuthbert
CI is currently failing because of a new cuthbert release. We should upgrade to v0.0.14 (and probably pin, as the cuthbert API is not super stable).

DanWaxman:
Cuthbert cut a new release which had some breaking changes (in particular, filters no longer call init_prepare themselves, that's now our jobs). There was also a changing in the filenaming for EnKF (mia culpa). This pins an exact version, since I think cuthbert is still undergoing breaking changes every now and then.

This should also resolve some of the errors in #298.


MatthieuDarcy (approved):
Seems good to me

PR #301, opened 2026-08-07
Add EnRTS Smoother

DanWaxman:
The EnRTS smoother is now available in v0.0.14 of cuthbert. This PR does the proper wiring to have this in dynestyx :)

Should be merged after #300


mattlevine22 (approved):
This looks great, thanks!

I asked Codex to scan for bugs, and its main complaint (can get NaN rollouts if n_particles<= state_dim) was actually a bit more general, which I turned into #303.

PR #305, opened 2026-08-10
Upgrade to jax>=0.8.2

MatthieuDarcy:


DanWaxman:
Does this not cause issues with `numpyro==0.21.0` and `jax>=0.10.0`?

MatthieuDarcy:
I think the issue is `numpyro<0.21.0`, with `numpyro==0.21.0` this is fine. I'm updating the dependencies to fix numpyro to that version.

DanWaxman:
Right, sorry, it was numpyro 0.20.1 that was broken. We should guard against that.

MatthieuDarcy:
Agreed, should be fixed now, it is now `jax>=0.8.2` and `numpyro >=0.21` (only version that allows for the latest jax version).

mattlevine22 (approved):
I'm good with this. @DanWaxman do you have any concerns? I don't really know how to be "sure" about things like this.

DanWaxman (approved):

DanWaxman:
Yeah I think if we're pinning `numpyro >= 0.21.0` we're okay.

PR #306, opened 2026-08-11
run test suite on main daily at 12 UTC

mattlevine22:
Hope is to catch simple build errors resulting in package shifts out in the universe.

There are other fancier things to do, but this seems simple and worthwhile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,8 @@
 name: Test
 
 on:
+  schedule:
+    - cron: "0 12 * * *"
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,9 @@ on:
       - '*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest


DanWaxman:
Great… though I am also somewhat scared of supply chain attacks, we should just be extra sure this is not running in an environment with any secrets.

DanWaxman (approved):

mattlevine22:
Do you know how to know?

mattlevine22:
Codex's security assessment: 

This scheduled workflow runs the existing test suite against the default branch and does not reference repository secrets. Its temporary GITHUB_TOKEN is explicitly limited to read-only repository access (contents: read). The remaining risk is the standard CI supply-chain exposure from third-party actions and installed dependencies; this change does not materially increase that risk.

** It suggested I add `permissions: read`, so I did that above.

mattlevine22:
I'd love for us to put time into shoring up these kinds of security vulnerabilities and creating processes that will scale well as we grow. Unfortunately, we don't have the resources/bandwidth to prioritize that currently....one of the many reasons we are applying for funding from the [NSF PESOSE call](https://www.nsf.gov/funding/opportunities/pesose-pathways-enable-secure-open-source-ecosystems/nsf26-506/solicitation)

Issue #304, opened 2026-08-10
Add support for IMEX solvers


MatthieuDarcy:
Diffrax supports IMEX schemes but dynestyx cannot pass a `MultiTerm` therefore preventing their use. #307

PR #307, opened 2026-08-11
Added support for multiterm (imex solvers)

MatthieuDarcy:
Requires `diffrax==0.7.2` which is only possible when CD-Dynamax is updated, so marking this as a draft in the meantime.
#304


dynestyx/models/checkers.py
```diff
@@ -103,14 +103,67 @@ def _make_probe_state(
...
 
 
+@runtime_checkable
+class _ImExDriftLike(Protocol):
+    """Structural check for ImExDrift-shaped drifts.
+
+    Checked structurally (not via `isinstance(..., ImExDrift)`) because the
+    concrete `ImExDrift` class lives in `state_evolution.py`, which imports
+    from `core.py`, which imports from this module -- importing it here
+    would cycle.
```
mattlevine22: This doesn't seem like a good reason to create a dummy class. Can you take a look to see what sort of file refactor / creation would be needed to avoid this circular import issue?

dynestyx/types.py
```diff
@@ -16,6 +17,88 @@ def __call__(
         raise NotImplementedError()
 
 
+type _Drift = Callable[
```
mattlevine22: How about `models/types.py`?

mattlevine22: just trying to think about big picture organization...`dynestyx/types.py` are pretty high-level types, so seems weird to drop ImExDrift there.

mattlevine22: actually, I guess `models/core.py` is already mostly a "types.py" file.

mattlevine22: How about `models/drifts.py` and include `Drift`, `AffineDrift`, and `ImExDrift` there?

MatthieuDarcy: I'm going to be a bit verbose to make sure I understand things correctly, let me know if I'm off track.

Right now in`checkers`.py, we use `_validate_continuous_state_evolution` to check that we are not using both an `ImExDrift` AND a `state_evolution.potential` term (because we decided that combining these two is ambiguous), raising an error if so.

I think adding a `drifts.py` with `Drift`, `AffineDrift`, `ImExDrift`, `Potential` (?) and importing them into `checkers.py` could make sense. 

One thing I'm not completely sure about is that `diffusion.py` is imported at the level of `core.py`, one level down so there is a difference between the two (and I would naively assume that they would mirror each other, I don't know if this is a concern). I could see if there is a way of checking the `ImExDrift` vs `Potential` at the level of `core.py` but it's definitely more complicated.

I want to add that I believe that we have 3 choices:

1. add the check in `checkers.py`, in `_validate_continuous_state_evolution`.
2. add the check in `core.py`, in `ContinuousTimeStateEvolution` (doing it during class instantiation(
3. add the check in `odes.py`, in `_solve`
I'm not a fan of 3, because the error will be raised much later.

mattlevine22: I think it makes sense to include Potential too!

Not sure I understand the concern about `diffusions.py`. Both `drifts.py` and `diffusions.py` would be leaf modules imported by core.py; and checkers.py would additionally import ImExDrift

That should be OK? want to understand the complication here

MatthieuDarcy: I'm just wanted to make sure that it was fine that `checkers.py` depends on `drifts.py` but not on `diffusions.py` is fine. I might be overthinking this, if it's not an issue, then will implement.

mattlevine22: yeah, `checkers.py` COULD depend on `diffusions.py` if we needed it to though right?

We just don't do any diffusion checks

MatthieuDarcy: Yes

MatthieuDarcy:
There now is a `drifts.py`, imported by `core.py` and `checkers.py`. It's a leaf, no internal Dynestyx imports. I think the solution works pretty well, let me know what you think @mattlevine22

docs/api_reference/public/models/core/drift.md
mattlevine22, line 1: I guess this file path should change to `models/drifts/drift.md`

docs/api_reference/public/models/core/potential.md
mattlevine22, line 1: I guess this file path should change to `models/drifts/potential.md`

docs/api_reference/public/models/specialized/affine_drift.md
mattlevine22, line 1: change file path

mattlevine22:
> There now is a `drifts.py`, imported by `core.py` and `checkers.py`. It's a leaf, no internal Dynestyx imports. I think the solution works pretty well, let me know what you think @mattlevine22

Yeah looks reasonable---I guess you need to change some of the docs .md file paths accordingly though

mattlevine22:
@MatthieuDarcy I did a quick iteration with Codex, which picked up 2 issues:

1. `MultiTerm` is not specific to IMEX. Diffrax also uses it for SDE and Langevin solvers, where the terms have different meanings. Checking only `get_origin(solver.term_structure) is dfx.MultiTerm` can therefore misclassify solvers like `EulerHeun` and pass them two ODE terms incorrectly.

   Recommended fix: require both `MultiTerm` and `dfx.AbstractImplicitSolver`, or use an explicit supported-solver check:

   ```python
   is_imex_solver = (
       isinstance(solver, dfx.AbstractImplicitSolver)
       and get_origin(solver.term_structure) is dfx.MultiTerm
   )
   ```

2. A potential drift can be silently discarded inside `dsx.plate`. The `ImExDrift + potential` validation is skipped in plate contexts, while the IMEX solve evaluates only `explicit_term` and `implicit_term`, bypassing `total_drift` and omitting the potential gradient.

   Recommended fix: perform the incompatibility check unconditionally, outside plate-skipped shape validation. Alternatively, have `ContinuousTimeStateEvolution` construct the complete IMEX decomposition, including the potential contribution. A plated regression test would be helpful.

mattlevine22:
Also, the reaction diffusion notebook kinda uses IMEX at the end, but it doesn't show any plots or justify the choice

MatthieuDarcy:
> @MatthieuDarcy I did a quick iteration with Codex, which picked up 2 issues:
> 
>     1. `MultiTerm` is not specific to IMEX. Diffrax also uses it for SDE and Langevin solvers, where the terms have different meanings. Checking only `get_origin(solver.term_structure) is dfx.MultiTerm` can therefore misclassify solvers like `EulerHeun` and pass them two ODE terms incorrectly.
>        Recommended fix: require both `MultiTerm` and `dfx.AbstractImplicitSolver`, or use an explicit supported-solver check:
>        ```python
>        is_imex_solver = (
>            isinstance(solver, dfx.AbstractImplicitSolver)
>            and get_origin(solver.term_structure) is dfx.MultiTerm
>        )
>        ```
> 
>     2. A potential drift can be silently discarded inside `dsx.plate`. The `ImExDrift + potential` validation is skipped in plate contexts, while the IMEX solve evaluates only `explicit_term` and `implicit_term`, bypassing `total_drift` and omitting the potential gradient.
>        Recommended fix: perform the incompatibility check unconditionally, outside plate-skipped shape validation. Alternatively, have `ContinuousTimeStateEvolution` construct the complete IMEX decomposition, including the potential contribution. A plated regression test would be helpful.

I fixed both.

1. <strike>In addition to what you suggested, I added a check at the very beginning of `solve_ode_state_path`  raising an error if the solver is of type`dfx.AbstractStratonovichSolver` or`dfx.AbstractItoSolver`. The reason for this is because it's technically possible to have _implicit SDE_ solvers, they just don't exist in diffrax yet. Checking for `dfx.AbstractImplicitSolver` only could therefore lead to a bug in a potential future release? In principle this should not alter `solve_ode_state_path` as it should already error if given an SDE solver and the error message should now be clearer. Let me know if you approve of this (another option is to add this check only when there is an `ImExDrift`, more specific).</strike> I put the `dfx.AbstractImplicitSolver` check only.
2. I added a specific check for `ImExDrift` and `Potential` in `checkers.py` , called outside the `_inside_plate` check, only checks for `continuous_time`.
Will work on the docs later.

mattlevine22:
Yeah looks good, thanks!

MatthieuDarcy:
Actually nevermind, ODE solvers can also be labelled as `dfx.AbstractStratonovichSolver` or `dfx.AbstractItoSolver`. I'm changing this to only check for `AbstractImplicitSolver`, might break in a future release but so be it.

DanWaxman:
FYI I updated the PR to merge to main + made a merge commit, since the previous base branch landed on main.

mattlevine22 (approved):
I'm happy with this PR, so approving it.

Would recommend updating the end of the notebook with a plot of the pde solution that used the imex solver though.

MatthieuDarcy:
@mattlevine22 I modified some of the API documentation in `/core`

The order is changed (it seemed a bit random, I tried to have more coherent order). 

There now is a dedicated `Drifts`section, similar to `Diffusions`, with `Drift`, `Potential`, `AffineDrift`, and `ImExDrift`. 
Note that `AffineDrift`has its own separate documentation in `Specialized`, I didn't want to touch it.
 
I updated the notebook with a plot and added a justification for using Imex.

Let me know what you think, didn't want to merge without your approval.

dynestyx/models/drifts.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    Drift vector field for continuous-time state evolution.
+
+    Mathematically, the drift is a mapping
+    $\\mu: \\mathbb{R}^{d_x} \\times \\mathbb{R}^{d_u} \\times \\mathbb{R}
+    \\to \\mathbb{R}^{d_x}$, i.e., $(x, u, t) \\mapsto \\mu(x, u, t)$.
+    In the SDE formulation used by `ContinuousTimeStateEvolution`,
+    $dx_t = \\mu(x_t, u_t, t) \\, dt + \\sigma(x_t, u_t, t) \\, dW_t$, this
+    mapping forms the $\\mu$ term.
+
```
DanWaxman, line 21: I think we've been using $f$ for drifts and $g$ for diffusions generally.

dynestyx/models/drifts.py
```diff
@@ -0,0 +1,200 @@
...
+    1. As a drop-in `drift` for `ContinuousTimeStateEvolution` used with an
+       ordinary explicit solver (e.g. `diffrax.Tsit5`) or a fully-implicit
+       solver (e.g. `diffrax.Kvaerno4`) -- `__call__` returns the combined
+       vector field $f(x, u, t) = f_{ex}(x, u, t) + f_{im}(x, u, t)$.
+    2. As the source of the explicit/implicit split consumed by diffrax's
+       IMEX solvers (e.g. `diffrax.KenCarp3/4/5`, `diffrax.Sil3`, which
+       require diffrax's `MultiTerm`), via `make_imex_tuple`.
+
+    Construct with keyword-only arguments to avoid silently swapping the two
+    terms: `ImExDrift(explicit_term=..., implicit_term=...)`.
```
DanWaxman:
```suggestion
    Must be constructed with keyword-only arguments to avoid silently swapping the two
    terms: `ImExDrift(explicit_term=..., implicit_term=...)`.
```

dynestyx/models/drifts.py
```diff
@@ -0,0 +1,200 @@
...
+        self.explicit_term = explicit_term
+        self.implicit_term = implicit_term
+
+    def __call__(
+        self,
+        x: Real[Array, " state_dim"] | Real[Array, ""],
+        u: Real[Array, " control_dim"] | Real[Array, ""] | None,
+        t: float | int | Real[Array, ""],
+    ) -> Real[Array, " state_dim"] | Real[Array, ""]:
+        return self.explicit_term(x, u, t) + self.implicit_term(x, u, t)
```
DanWaxman, line 184: Wouldn't this be where a warning would go?

dynestyx/models/state_evolution.py
```diff
@@ -276,44 +275,3 @@ def __call__(self, x, u, t_now, t_next):
...
-
-    This is commonly used as the drift term $\\mu(x_t, u_t, t)$ inside
-    `ContinuousTimeStateEvolution`, and is a building block for LTI models such as
-    `LTI_continuous`.
-
-    Attributes:
-        A (jax.Array): Drift matrix with shape $(d_x, d_x)$.
-        B (jax.Array | None): Optional control matrix with shape $(d_x, d_u)$.
-        b (jax.Array | None): Optional additive bias with shape $(d_x,)$.
-    """
```
DanWaxman, line 301: I think we have enough (even if the total number is small) users and this is a common enough class that we should be keeping this API entrypoint with a deprecation warning that it will be removed in `v0.5.0`.

DanWaxman (changes requested):
Some small comments

DanWaxman:
Sorry I realized afterwards that Matt had approved it already lol. But these are tiny, I'm happy for it to be merged afterwards :)

dynestyx/models/drifts.py
MatthieuDarcy: The reason I didn't put it here (it's in `solve_ode_state_path`) is because I wanted to tell the user which solver they were calling. It's also where an error is thrown if the solver requires `ImExDrift` so it felt natural to have these next to each other.

dynestyx/models/drifts.py
MatthieuDarcy: It's not really consistent. In the [SDE tutorial](https://basisresearch.github.io/dynestyx/stable/tutorials/gentle_intro/06_continuous_time/) and the API, $L$ is used for diffusion. $\mu$ is used in the API for drift but not in the tutorials. 

Lets pick a convention and stick to it? I can make the changes in the API.

dynestyx/models/state_evolution.py
MatthieuDarcy: `state_evolution.py` now imports `AffineDrift` from `drifts.py`so this restores the API entrypoint (with a warning).

dynestyx/models/drifts.py
DanWaxman: Fair enough

dynestyx/models/drifts.py
DanWaxman: Fair. We can do this later. I made https://github.com/BasisResearch/dynestyx/issues/327.

DanWaxman (approved):
Thanks!!

PR #315, opened 2026-08-13
Bump Version to 0.3.0

DanWaxman:


mattlevine22 (approved):
Looks good, I made it 0.3.0---I like the idea of using these middle-version-numbers to keep users alert to things that may change their life (here some changes in required dependencies). Ideally, the last number shouldn't change their existing usage (other than "improving it")

PR #319, opened 2026-08-14
Bump version to 0.3.1

mattlevine22:
New closed loop controlled simulator interface


DanWaxman (approved):
the matthieu release is coming

PR #320, opened 2026-08-14
Upgrade diffrax + cd-dynamax

DanWaxman:
Unblocks #304 by upgrading diffrax version.


mattlevine22 (approved):
forgot to approve this, works locally, its HOTTOGO

PR #321, opened 2026-08-14
Observation scoring PR

mattlevine22:
Supersedes #296


dynestyx/evaluation/observation_scoring.py
```diff
@@ -0,0 +1,261 @@
...
+                    sample_seed=scoring_config.sample_seed,
+                )
+            else:
+                raise NotImplementedError(
+                    f"Unsupported observation scoring rule type: {type(rule).__name__}."
+                )
+        except NotImplementedError:
+            if scoring_config.unsupported == "skip":
+                continue
+            raise
```
DanWaxman: I don't fully get this API choice, i.e., to allow for unsupported scores to be requested.

dynestyx/inference/observation_predictions.py
DanWaxman, line 1: Not really a comment about anything in particular, but I'm surprised this is such a big file!

DanWaxman (changes requested):
Few small comments. I also opened #323 for some other changes.

In general, I find it a bit sad we can't use this with `LatentPathBuilder`. I guess this is fine as an intermediate draft so we get the feature out, but I'd like to have a more general `Simulator`-based inference interface here. This should enable use with LPB _and_ scoring at lead time. But again happy to boot that to some other time in the near future.

dynestyx/inference/observation_predictions.py
mattlevine22: Reduced from 265 -> 215

dynestyx/evaluation/observation_scoring.py
mattlevine22: removed

DanWaxman (approved):
👍

PR #322, opened 2026-08-16
Add an adaptive Metropolis-within-Gibbs sampler

DanWaxman:
This adds an adaptive Metropolis-within-Gibbs kernel, inspired by the kernel in PFJax. It additionally adds some room for diagnostics, and a quick separate fix to change Simulator's `throw` setting. The kernel works very well in low-dimensional and simple problems, though has some sensititiity to initialization.


dynestyx/inference/configs/mcmc.py
```diff
@@ -53,7 +55,65 @@ class HMCConfig(BaseMCMCConfig):
...
+    One transition updates each flattened unconstrained coordinate in order.
+    Proposal scales adapt during warmup toward the requested acceptance rate
+    and remain fixed while retained samples are generated.
+
+    This sampler is currently implemented by the BlackJAX integration only.
+
+    Attributes:
+        initial_proposal_scale (ArrayLike): Positive scalar proposal scale, or
+            one positive scale per flattened unconstrained coordinate.
+        target_acceptance_rate (float): Per-coordinate target acceptance rate.
```
mattlevine22: probably not "per-coordinate" right?

dynestyx/inference/configs/mcmc.py
```diff
@@ -2,6 +2,8 @@
 from collections.abc import Callable
 from typing import Literal
 
+import numpy as np
```
mattlevine22: could we use jnp?

dynestyx/inference/configs/simulator.py
```diff
@@ -33,6 +33,9 @@ class ODESimulatorConfig:
             this is the fixed integration step. Defaults to `1e-3`.
         max_steps (int): Maximum number of integration steps permitted by
             `diffrax.diffeqsolve`. Defaults to `100_000`.
+        throw (bool): Whether Diffrax raises when a solve fails. Set this to
```
mattlevine22: oooo I think you knew I would like this :)

dynestyx/inference/configs/mcmc.py
```diff
@@ -53,7 +55,65 @@ class HMCConfig(BaseMCMCConfig):
...
+    target_acceptance_rate: float = 0.8
+
+    def __post_init__(self) -> None:
+        if not 0.0 < self.target_acceptance_rate < 1.0:
+            raise ValueError("target_acceptance_rate must be between 0 and 1")
+
+
+@dataclasses.dataclass
+class AdaptiveMetropolisConfig(BaseMCMCConfig):
+    """Adaptive componentwise random-walk Metropolis configuration.
```
mattlevine22: should clarify that it is a MVN proposal distribution

dynestyx/inference/integrations/blackjax/adaptive_metropolis.py
```diff
@@ -0,0 +1,164 @@
...
+# https://github.com/mlysy/pfjax/blob/97652aa1bdff73a92c0286549b010e99cc6f7264/src/pfjax/mcmc.py
+
+from collections.abc import Callable
+from typing import NamedTuple
+
+import blackjax
+import jax
+import jax.numpy as jnp
+import jax.random as jr
+from blackjax.base import SamplingAlgorithm, build_sampling_algorithm
```
mattlevine22, line 13: I failed a test here locally after `sync --dev --all-extras`

Codex says `build_sampling_algorithm` was added in BlackJAX 1.6

mattlevine22 (approved):
This looks great to me, thanks Dan!

mattlevine22:
Shoot--- @DanWaxman I didn't mean to accept (because tests fail locally with old blackjax), although if you fix the above I'm totally happy with you merging

dynestyx/inference/configs/mcmc.py
DanWaxman: It's kinda per-coordinate. It's a Metreopolis-within-Gibbs, which I'm realizing was not documented as clearly as I would have wanted. So it's the target acceptance rate, for each coordinate's Metropolis sampler.

dynestyx/inference/configs/mcmc.py
DanWaxman: Yup!

dynestyx/inference/configs/mcmc.py
DanWaxman: As above, it's a Metropolis-within-Gibbs, which was pretty poorly documented. I've fixed this, and clarified they're per-coordinate scalar normal proposals.

dynestyx/inference/integrations/blackjax/adaptive_metropolis.py
DanWaxman: I updated the minimum required version.

DanWaxman:
I pushed a new commit to resolve the above. I also realize it was poorly documented/communicated this is a Metropolis-within-Gibbs, so it's running Metropolis steps per-coordinate within a Gibbs sampler. Documentation should be better for this now. If you could please take a look and merge :)

mattlevine22 (approved):
great thanks for the updates---approving for real now!

PR #325, opened 2026-08-18
Add an ODE discretizer

### LINKED ISSUE #324: Add discretizer support for ODEs
1. Add a discretizer that takes Deterministic ContinuousTimeStateEvolution and maps it to a Dirac delta `x_{k+1} ~ Dirac(F(x_k))`, where `F` is a specified ODE solver.
- show that this allows us to use cuthbert filtering (perhaps with covariance inflation) on L63 ODE data

2. Consider a discretizer (or "relaxer") that takes Deterministic ContinuousTimeStateEvolution and instead maps it to a Gaussian transition `x_{k+1} = N(F(x_k); Sigma), where `F` is a specified ODE solver.

The goal would be to be able to efficiently handle state and parameter estimation for data from a deterministic ODE (e.g., L63) using our current filtering/MCMC stack.

This would be a continuation of the work in #299.

DanWaxman:
go with the flow


mattlevine22:
Looks great!

I found that setting `jitter = 0.1` for EKF made it work just fine for filtering (didn't bother with parameter estimation)---mind if I push that change to the notebook?

mattlevine22 (approved):
Looks good! I pushed a change to the notebook (uses jitter=0.1 for EKF so that it works well). 

Feel free to accept that or revert it if you really want to show EKF failing with bad choice of jitter. 

Approving either way to keep the flows flowing

DanWaxman:
> I found that setting jitter = 0.1 for EKF made it work just fine for filtering --- mind if I push that change to the notebook?

sure, that sounds good

> (didn't bother with parameter estimation)

right, this is harder because of the PID controller thing, things take forever with constant step size.

will merge after tests are green

Issue #312, opened 2026-08-12
Define observation/control alignment for discrete-time models


mattlevine22:
Discrete-time control currently has two possible conventions:

- `same_time`: \(y_k\) uses \(u_k\). This is the current `dsx.simulate` convention.
- `previous_transition`: \(y_0\) uses no control and \(y_{k+1}\) uses \(u_k\), the control that produced \(x_{k+1}\). This is the closed-loop simulator convention.

MPPI uses `dsx.simulate` internally, so observation-dependent losses may be off by one relative to the closed-loop system when observations depend on controls.

### Recommendation

Make the convention an explicit property of discrete-time `DynamicalModel`s:

```python
observation_control_alignment: Literal[
    "same_time",
    "previous_transition",
] = "same_time"
```

All consumers—simulation, filtering, smoothing, and closed-loop control—should use this model-level contract consistently. Preserve `"same_time"` as the default for backward compatibility.

MatthieuDarcy:
so dsx.simulate implements the following loop

$$
\begin{aligned}
x_0 &\sim p_0 \\
y_0 &\sim p(\cdot |x_0, u_0)
\end{aligned}
$$

and subsequently

$$
\begin{aligned}
x_{k+1} &\sim p(\cdot | x_k, u_k)\\
y_k &\sim p(\cdot| x_k, u_k)
\end{aligned}
$$

but MPPI/MPC really wants

$$
\begin{aligned}
x_0 &\sim p_{\text{Filter}}\\
x_{k+1} &\sim p(\cdot | x_k, u_k) \\
y_{k+1} &\sim p(\cdot| x_{k+1}, u_k)
\end{aligned}
$$

where $p_{\text{Filter}}$ is the filtering distribution and $y_0$ _has already been observed_ (the 0 is really the last time step from the online control loop).

Right now MPPI solves this by duplicating the last control which means that

$$
y_0 \sim p(\cdot| x_0, u_0)
$$

Even though $y_0$ already exists. This means that if the observation distribution depends on $u$ and the loss depends on $y_0$, then this spurious term could affect the chosen control.

All this to say that I believe that if `"previous_transition"` is chosen, then $y_0$ is never sampled and the `SimulatedResult` should return $x_1, y_1, \dots$ (no initial condition/observation) which would have the benefit of preserving the correspondence between number of controls and number of states/observations.

Issue #272, opened 2026-07-14
Enforce universal consistency of time-handling


mattlevine22:
There are a number of partially-interrelated issues surrounding the handling/interpretation of times in the repo. 

Key considerations are how to interpret `obs_times` and `ctrl_times` consistently across discrete-time and continuous-time uses.

Will add sub-issues to this meta issue.

Issue #96, opened 2026-02-13
Add support for flexible control `u` interpolation schemes


mattlevine22:
May require changes to CD-Dynamax and/or our plumbing to it.

PR #331, opened 2026-08-19
Define observation/control alignment for discrete-time models

MatthieuDarcy:
Add observation_control_alignment for discrete-time Simulator (#312).

# Mathematical summary
Implements the `same_time` vs `previous time distinction`. `same_time` (default, preserves the existing behavior) implements:

$$
\begin{aligned}
x_0 &\sim p_0 \\
y_0 &\sim p(\cdot |x_0, u_0)
\end{aligned}
$$

and subsequently

$$
\begin{aligned}
x_{k+1} &\sim p(\cdot | x_k, u_k)\\
y_k &\sim p(\cdot| x_k, u_k)
\end{aligned}
$$

`previous_time` implements

$$
\begin{aligned}
x_0 &\sim p_{0}\\
x_{k+1} &\sim p(\cdot | x_k, u_k) \\
y_{k+1} &\sim p(\cdot| x_{k+1}, u_k)
\end{aligned}
$$

# Summary of changes

## `previous_transition` contract

Adds an explicit property of `DynamicalModel` defined at initialization called `observation_control_alignment: Literal[
            "same_time", "previous_transition"
        ] = "same_time"`

When defined as `previous_transition`, this results in the following behavior

1. $y_0$ cannot be sampled. Hence observations $y$ is of size $T-1$ while the states $x$ are of size $T$.
2. I added the controls $u$ in the returned`SimulatedResults`. This is very practical when doing MPC. This leads to a similar behavior as observations: when using "same_time" it is of size $T$, when using "previous_time" it is of size $T-1$. Only available when running `DiscreteTimeSimulator`.

This means that `states` and `times` are of different sizes to `controls` and `observations` (specifically size $T$ vs $T-1$) when using `previous_transition`.

1. MPPI is updated to use this convention in #348

## Updated `DiscreteControlLoopSimulator` 

`DiscreteControlLoopSimulator` now deliberately always uses the previous-transition convention, independent of `dynamics.observation_control_alignment`:

$$
\begin{aligned}
x_0 &\sim p_0, \qquad \hat p_0 = p_0, \\
(u_k, s_{k+1}) &= \pi(\hat p_k, t_k, t_{k+1}, s_k), \\
x_{k+1} &\sim p(\cdot \mid x_k, u_k, t_k, t_{k+1}), \\
y_{k+1} &\sim p(\cdot \mid x_{k+1}, u_k, t_{k+1}).
\end{aligned}
$$

Previously, it was assumed that the observation function could support `u=None`. A closed-loop result therefore has $T$ states, times, and beliefs, and $T-1$ aligned observations and controls. This matches the `previous_transition` convention and avoids imposing an implicit requirement that the observation model accept `u=None` or be control-independent, but is change in the behavior. 

# Future work

1. `previous_transition` only works for simulation, no conditioning and not filtering. Worth delegating to a seperate PR?
2. Adding controls in `SimulatedResults` for ODE/SDE/Continuous time. It might also be worthwile to expand support for continuous control beyond zero-order hold.
3. `LTI_discrete` does not accept `observation_control_alignment` as a keyword, I did not touch it but worth looking at it once we are satisfied with the contract.


mattlevine22:
"DiscreteControlLoopSimulator now deliberately always uses the previous-transition convention, independent of dynamics.observation_control_alignment:"
- Hmm, seems better to keep this aligned with reality somehow. I suppose if `observation_control_alignment` is not specified by a user, then we are free to choose the "right" default; if a user does specify it, we should error if they made an incompatible choice.

mattlevine22:
"previous_transition only works for simulation, no conditioning and not filtering. Worth delegating to a seperate PR?"
- Yes, that seems like the right way to do it. Could you create an issue and then make it a sub-issue of #272?

mattlevine22:
"Adding controls in SimulatedResults for ODE/SDE/Continuous time. It might also be worthwile to expand support for continuous control beyond zero-order hold."
- Yes, this is basically #96

docs/api_reference/public/control.md
```diff
@@ -6,22 +6,22 @@ single discrete-time trajectory. At each step it performs
 \[
```
mattlevine22: This isn't rendering on the GitHub or cursor markdown viewers. Mind massaging this somehow to get it to appear? sometimes you need to wrap long latex lines in `...`, not really sure though.

docs/api_reference/public/control.md
```diff
@@ -6,22 +6,22 @@ single discrete-time trajectory. At each step it performs
...
-state. This differs from the same-index convention used for a precomputed
-open-loop control trajectory. This is a temporary difference: [Issue
-#312](https://github.com/BasisResearch/dynestyx/issues/312) tracks aligning
-closed-loop control with the simulator convention and requiring controlled
-`DynamicalModel` observation models to follow that convention.
+The initial policy decision uses the model's initial-state distribution as its
+belief; no synthetic initial observation is generated. Every observation at
+`t[k + 1]` receives `u[k]`, the control that produced its state. Closed-loop
+simulation therefore always follows the `"previous_transition"` convention,
+independently of `dynamics.observation_control_alignment`. For `T` prediction
```
mattlevine22, line 22: Would rather us give an error if the user selects the wrong `dynamics.observation_control_alignment`. We should allow them to omit it and then default to the appropriate choice at the right time...hopefully not easier said than done?

docs/tutorials/control/mpc_demo.ipynb
```diff
@@ -307,7 +308,7 @@
     "for result, label, color in runs_mppi:\n",
     "    t = result.times[0]\n",
     "    filtered_mean = result.filtered_states_mean[0]\n",
-    "    axes[0].plot(t, result.observations[0, :, 0], \".\", color=color, alpha=0.4, label=f\"{label} (observed)\")\n",
+    "    axes[0].plot(t[1:], result.observations[0, :, 0], \".\", color=color, alpha=0.4, label=f\"{label} (observed)\")\n",
```
mattlevine22, line 311: I don't love that the user has to know to do `t[1:]`. I've messed stuff like that up so many times over the years.

What do you think about having explicit fields for the observation times, state times, and control times?

@DanWaxman I know this may be a bit wasteful, but (a) maybe it can be optional to return them and (b) I really think ease of use here may be worth it. At least times are scalar so it really just scales with length of the series (not dimensionality of the system).

docs/tutorials/control/mpc_demo.ipynb
```diff
@@ -11,15 +11,16 @@
     "$$\n",
     "\\begin{aligned}\n",
     "&x_0 \\sim p(x_0)\\\\\n",
-    "&y_0 | x_0 \\sim p(y_0 | x_0, t_0) \\\\\n",
-    "&\\hat{x}_{0|0} = \\text{FilterUpdate}(y_0, t_0) \\\\\n",
```
mattlevine22, line 15: Should still define all the `0` variables so that this system of relations is fully specified. I'll try to flag that where I see it, but maybe ask uncle AI to find those spots.

dynestyx/models/core.py
```diff
@@ -141,12 +150,19 @@ def __init__(
         observation_dim: int | None = None,
         categorical_state: bool | None = None,
         continuous_time: bool | None = None,
+        observation_control_alignment: Literal[
```
mattlevine22, line 153: After thinking on it, perhaps for now leave this as is with `same_time` default; When doing control stuff, if it is `same_time`, issue a warning that you are changing it to previous.

dynestyx/simulation/discrete.py
```diff
@@ -273,23 +303,36 @@ def _sim_one_trajectory(
...
         states, observations = jax.vmap(_sim_one_trajectory)(sim_keys, initial_state)
+
+        controls = None
+        if ctrl_values is not None:
+            controls = _ensure_trailing_dim(
+                jnp.broadcast_to(ctrl_values[None], (n_sim, *ctrl_values.shape))
+            )
+
         return SimulatedResult(
             times=_tile_times(times, n_sim),
```
mattlevine22, line 331: In reference to my earlier comment about associating times with states/observations/controls...how about just storing the first time index for states/observations/controls here?

tstart_idx_states, tstart_idx_observations, tstart_idx_controls? or something like that

dynestyx/simulation/discrete.py
```diff
@@ -315,8 +358,13 @@ def simulate(
         if predict_times is None:
             raise ValueError("predict_times must be provided")
 
+        align_times = (
+            predict_times[:-1]
```
mattlevine22, line 362: Hmmm I'm kinda questioning this, though possibly a rabbit-hole connected to #272.

Am I understanding correctly that in "previous" mode, `predict_times[0]` is used to sample `x0` and nothing else?

In that case, I would much prefer that the user simply specify `t0` directly, and then omit that value from `predict_times`.

Overall, I worry it is confusing to specify a vector of predicted times and get 1 less back.

tests/test_hierarchical_simulator_discretizer_smokes.py
```diff
@@ -275,6 +309,25 @@ def test_plate_forward_discrete_ode_sde_shapes(source):
...
+            _plate_previous_transition_model(
+                predict_times=predict_times,
+                ctrl_times=ctrl_times,
+                ctrl_values=ctrl_values,
+            )
+
+    assert tr["f_times"]["value"].shape == (2, 1, 4)
+    assert tr["f_states"]["value"].shape == (2, 1, 4, 2)
+    assert tr["f_observations"]["value"].shape == (2, 1, 3, 1)
+    assert tr["f_controls"]["value"].shape == (2, 1, 3, 1)
```
mattlevine22, line 328: I would assert that things are not NaNs. That has bitten us before.

tests/test_discrete_control.py
```diff
@@ -663,10 +660,91 @@ def model():
...
+
+    result = dsx.simulate(
+        dynamics,
+        rng_key=jr.PRNGKey(0),
+        predict_times=predict_times,
+        control_policy=_simple_policy(),
+        filter_config=EKFConfig(),
+    )
+
+    assert result.states is not None
```
mattlevine22, line 697: I'd assert more clearly what these should be, e.g. finite arrays

tests/test_discrete_control.py
```diff
@@ -663,10 +660,91 @@ def model():
...
+def test_single_timepoint_closed_loop_returns_no_observations_or_controls():
+    result = dsx.simulate(
+        _lti_1d(),
+        rng_key=jr.PRNGKey(0),
+        predict_times=jnp.array([0.0]),
+        control_policy=_simple_policy(),
+        filter_config=KFConfig(filter_source="cuthbert"),
+    )
+
+    assert result.states is not None
```
mattlevine22, line 742: again, maybe assert these are finite arrays

mattlevine22: might even be worth a helper that is `assert_finite(array, shape=Optional)`

mattlevine22 (changes requested):
Overall, seems solid to me.

Main asks are:
1. Clean up the latex math and make sure it renders in markdown (I didn't read all of it because of that) and is fully specified including all zero cases (e.g. s_0 (even with a note that pi auto-initializes s0 in practice))

2. Try to return useful time-indexes that offer some redundancy to users in SimulatedResult (rather than needing to write things like `t[1:]`)

3. We should always at least issue a warning if we override a user's choice; here w.r.t. `dynamics.observation_control_alignment`.

4. Look to see if it is easy for the user to simply supply `t0` instead of appending it to `predict_times` in previous-mode; my guess is not, which is fine. May just be better dealt all at once with #272. Would be great to get your thoughts on this though.

5. Add some `is_finite` checks to the tests (esp hierarchical)...we've gotten fooled in the past and had passing tests that were producing NaNs.

NB:  there are a lot of new tests, which we've been resistant of. However, I think they seem reasonable, and in particular, when messing with indexing I feel better about more guardrails and edge case tests, so I'm happy with these.

MatthieuDarcy:
> "DiscreteControlLoopSimulator now deliberately always uses the previous-transition convention, independent of dynamics.observation_control_alignment:"
> 
>     * Hmm, seems better to keep this aligned with reality somehow. I suppose if `observation_control_alignment` is not specified by a user, then we are free to choose the "right" default; if a user does specify it, we should error if they made an incompatible choice.

After thinking about this, I think that we can support both conventions for online control (and keep the default `same_time`) at the cost of deviating from the existing convention we  introduced. Here is my summary:

### `same_time` convention
Notation:
-  $`\tilde{p}_k \approx p(x_k \mid y_0, \dots, y_{k-1},\ u_0, \dots, u_{k-1})`$is the predicted distribution.
- $\hat{p}_k \approx p(x_k \mid y_0, \dots, y_k,\ u_0, \dots, u_k)$ is the filtered distribution.

In this convention, the control $u_k$ drives the transition into the next state $x_{k+1}$ and generates the observation $y_k$ (they are aligned).

$$
\begin{aligned}
&x_0 \sim p_0, \quad \tilde{p}_0 = p_0 \\
&u_k = \pi(\tilde{p}_k) \\
&y_k \sim p(y_k \mid x_k, u_k) \\
&\hat{p}_k = \text{FilterUpdate}(y_k, \tilde{p}_k, u_k) \\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k) \\
&\tilde{p}_{k+1} = \text{PredictionUpdate}(\hat{p}_k, u_k)
\end{aligned}
$$

In this case, the number of observations matches the number of states.

### Under `previous_transition` convention
This is the closest to what we decided originally, but doesn't align with Dynestyx default for offline control.

- $`\tilde{p}_k \approx p(x_k \mid y_1, \dots, y_{k-1},\ u_0, \dots, u_{k-1})`$ is the predicted distribution.
- $`\hat{p}_k \approx p(x_k \mid y_1, \dots, y_k,\ u_0, \dots, u_{k-1})`$ is the filtered distribution.

In this convention, the control $u_k$ drives the transition into the next state $x_{k+1}$ and generates the observation $y_{k+1}$. Hence, $y_0$ never exists.

$$
\begin{aligned}
&x_0 \sim p_0, \quad \tilde{p}_0 = \hat{p}_0 = p_0 \\
&u_k = \pi(\hat{p}_k) \\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k) \\
&\tilde{p}_{k+1} = \text{PredictionUpdate}(\hat{p}_k, u_k) \\
&y_{k+1} \sim p(y_{k+1} \mid x_{k+1}, u_k) \\
&\hat{p}_{k+1} = \text{FilterUpdate}(y_{k+1}, \tilde{p}_{k+1}, u_k)
\end{aligned}
$$

Here, the number of observations does not match the number of states, but it does match the number of controls. The number of controls is exactly the number of transitions (i.e., one less than the number of states).

I made a drawing because this stuff can be confusing.

Online control convention excalidraw

Thoughts?

mattlevine22:
What you wrote above looks right to me and sounds consistent with this PR.

Sounds like the only update is that you are proposing to be more accommodating to time-alignment in the closed loop controller?

Seems good to me.

docs/api_reference/public/control.md
```diff
@@ -3,25 +3,25 @@
...
-y_0 \mid x_0 &\sim p(y_0 \mid x_0, t_0), \\
-\hat{x}_{0\mid0} &= \operatorname{FilterUpdate}(y_0, t_0), \\
-(u_k, s_{k+1}) &= \pi(\hat{x}_{k\mid k}, t_k, t_{k+1}, s_k), \\
+\hat p_0 &= p(x_0), \\
+(u_k, s_{k+1}) &= \pi(\hat p_k, t_k, t_{k+1}, s_k), \\
 x_{k+1} \mid x_k,u_k &\sim p(x_{k+1}\mid x_k,u_k,t_k,t_{k+1}), \\
 y_{k+1} \mid x_{k+1},u_k &\sim p(y_{k+1}\mid x_{k+1},u_k,t_{k+1}), \\
-\hat{x}_{k+1\mid k+1} &= \operatorname{FilterUpdate}
-  (\hat{x}_{k\mid k},u_k,y_{k+1},t_k,t_{k+1}).
+\hat p_{k+1} &= \operatorname{FilterUpdate}
```
mattlevine22: RIP now it gets mad: "The following macros are not allowed: operatorname"

MatthieuDarcy:
# Summary of changes

Open-loop control now supports both the `same_time` and `previous_transition` conventions (see below for a definition of both). `DiscreteControlLoopSimulator` (closed-loop control) now supports the `previous_transition` convention and uses it by default, warning the user if they do not specify the model's `observation_control_alignment` and raising an error if they specify `same_time`. There are some difficulties in supporting `same_time`.

`SimulatedResult` now has `obs_times` and `ctrl_times` fields, which are populated when relevant (at the suggestion of @mattlevine22). This has a small cost (we have to store these) but makes the results more useful and eases some of the technicalities (see below).

A unified principle is that `states` and `times` should align: no giving $N$ times and getting $N-1$ states (but overvations/controls might differ). 

I have simplified and reworked some of the tests to properly verify the correctness of the different conventions.

## In open-loop control

`same_time`

$$
\begin{aligned}
& \text{Times} = [t_0, \dots, t_{N}]\\
&\text{Controls} = [u_0, \dots, u_N]\\
&x_0 \sim p_0\\
&\text{for $k = 0,\dots N-1$}:\\
&y_{k} \sim p(y_k \mid x_k, u_k, t_k) \quad \text{emit observation}\\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k, t_k, t_{k+1}) \quad \text{State transition}\\
&y_{N} \sim p(y_N \mid x_{N}, u_{N}, t_N) \quad \text{Final observation}
\end{aligned}
$$

Here there are $N+1$ states, observations and controls, all living on the grid $`[t_0, \dots, t_N]`$. The final control $`u_N`$ is only ever used for the observation.

`previous_transition`

$$
\begin{aligned}
& \text{Times} = [t_0, \dots, t_{N}]\\
&\text{Controls} = [u_0, \dots, u_{N-1}]\\
&x_0 \sim p_0\\
&\text{for $k = 0,\dots N-1$}:\\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k, t_k, t_{k+1}) \quad \text{State transition}\\
&y_{k+1} \sim p(y_{k+1} \mid x_{k+1}, u_k, t_{k+1}) \quad \text{emit observation}
\end{aligned}
$$

We have $N+1$ states (on the time grid $`[t_0, \dots, t_N]`$).
We have $N$ controls (on the time grid $`[t_0, \dots, t_{N-1}]`$).
We have $N$ observations (on the time grid $`[t_1, \dots, t_N]`$).

## In closed-loop control

`same_time` convention

- $`\tilde{p}_k \approx p(x_k \mid y_0, \dots, y_{k-1},\ u_0, \dots, u_{k-1})`$ is the predicted distribution.
- $`\hat{p}_k \approx p(x_k \mid y_0, \dots, y_k,\ u_0, \dots, u_k)`$ is the filtered distribution.

The control $`u_k`$ drives the transition into the next state $`x_{k+1}`$ and generates the observation $`y_k`$ (they are aligned).

$$
\begin{aligned}
& \text{Times} = [t_0, \dots, t_{N}]\\
&x_0 \sim p_0, \quad \tilde{p}_0 = p_0 \quad \text{Initialization step} \\
&\text{for $k = 0,\dots N-1$}:\\
&u_{k}, s_{k+1} =\pi(\tilde{p}_{k}, t_{k}, t_{k+1}, s_k) \quad \text{Select the control}\\
&y_{k} \sim p(y_k \mid x_k, u_k, t_k) \quad \text{emit observation}\\
&\hat{p}_k = \text{FilterAnalysis}(y_k, \tilde{p}_k, u_k) \quad \text{Update the filtering distribution using the observation} \\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k, t_k, t_{k+1}) \quad \text{State transition} \\
&\tilde{p}_{k+1} = \text{PredictionUpdate}(\hat{p}_k, u_k) \quad \text{Predict the filtering distribution}
\end{aligned}
$$

Here the policy uses the *predicted* distribution.
We have $N+1$ states (on the time grid $`[t_0, \dots, t_N]`$).
We have $N$ controls (on the time grid $`[t_0, \dots, t_{N-1}]`$); we can never produce a control at time $`t_N`$.
We have $N$ observations (on the time grid $`[t_0, \dots, t_{N-1}]`$); we can never produce an observation at time $`t_N`$, because $`u_N`$ would need $`t_{N+1}`$.

**This is not implemented** because Cuthbert's `Filter` does not expose separate `predict` and `filter_update` steps. This seems possible, but I did not want to overload this PR. I added dummy functions that show how this could be done, but this will raise an error if a user tries to do this.

`previous_transition` (the default, with a warning)

- $`\hat{p}_k \approx p(x_k \mid y_1, \dots, y_k,\ u_0, \dots, u_{k-1})`$ is the filtered distribution.

$$
\begin{aligned}
& \text{Times} = [t_0, \dots, t_{N}]\\
&x_0 \sim p_0, \quad \hat{p}_0 = p_0 \quad \text{Initialization step} \\
&\text{for $k = 0,\dots N-1$}:\\
&u_{k}, s_{k+1} =\pi(\hat{p}_{k}, t_{k}, t_{k+1}, s_k) \quad \text{Select the control}\\
&x_{k+1} \sim p(x_{k+1} \mid x_k, u_k, t_k, t_{k+1}) \quad \text{State transition} \\
&y_{k+1} \sim p(y_{k+1} \mid x_{k+1}, u_k, t_{k+1}) \quad \text{emit observation}\\
&\hat{p}_{k+1} = \text{FilterUpdate}(y_{k+1}, \hat{p}_k, u_k) \quad \text{Update the filtering distribution using the observation}
\end{aligned}
$$

Here the policy uses the *filtering* distribution. In this convention, $`y_0`$ never exists.
We have $N+1$ states (on the time grid $`[t_0, \dots, t_N]`$).
We have $N$ controls (on the time grid $`[t_0, \dots, t_{N-1}]`$); we can never produce a control at time $`t_N`$.
We have $N$ observations (on the time grid $`[t_1, \dots, t_N]`$). We do not produce an observation at time $`t_0`$, because $`u_0`$ is used for $`x_1`$ and $`y_1`$.

image



# Sharp edges 

The new `previous_transition` convention removes the artificial first observation $y_0$ (which used no control). However as a result, the following simple model becomes somewhat tricky to implement

image

I think one way of doing this could be to add an explicit keyword `observe_y0` and then implement the following loop:

image


@mattlevine22 can you let me know if you think this makes sense? (also I tried fixing all the rendering issues).

dynestyx/simulation/discrete.py
MatthieuDarcy: I opted to return explicit `obs_times`, `ctrl_times` fields because sometimes you don't have the first observation, sometimes you don't have the last.

dynestyx/simulation/discrete.py
MatthieuDarcy: The mismatch is now fixed: now `times`/`predict_times` always aligns with `states`. I agree that this was very confusing.

tests/test_hierarchical_simulator_discretizer_smokes.py
MatthieuDarcy: Added a finiteness check

mattlevine22:
Is `same_time` the default for open-loop?

MatthieuDarcy:
Yes, to preserve the old behavior.

mattlevine22 (commented):
Overall, my recommendation is to accept this PR as is and move on.
- I don't think it breaks any existing stuff...and if we discover that it does, we will just fix it.
- `observe_y0` boolean property of `DynamicalModel` is a good idea, but better as a new PR. I think this PR is about as big as I can generally handle reviewing haha.

Notes:
- I suppose that IF you set `observe_y0=True`, then there should be a warning that says: "y0 will be generated with a 0 control value (ignore this warning if your observation model does not depend on control inputs". 
- Doing `observe_y0` will also require dealing with more branches of the code (e.g., asserting it is True in all the Filters/Smoothers/OpenLoopSimulators unless you want to start supporting it True...which is even more code).
- `DynamicalModel` has a `t0` property...is this being considered at all? It is a bit dicey, and needs to become a more powerful grounding property soon (perhaps not in this PR other than maintaining current internal consistency).

@DanWaxman could you please look at this briefly and decide whether you'd like us to (a) do some more careful reviewing or (b) go for it?

I read the code in detail in my last iteration, but now it is a bit tough for me to follow what has changed (@MatthieuDarcy this is an example where fewer targeted commits helps a lot in the review process)

MatthieuDarcy:
Yes my apologies for this, I know this is too much but it turned out to be a bit of a rabbit hole.

@mattlevine22 @DanWaxman  IF you want a more careful review, I can break it up into smaller PRs, doing one for open loop and one for closed loop (and removing some of the superfluous elements), but this will take me a bit of time.

mattlevine22:
My gut is that this is good enough but want to get Dan's take

docs/tutorials/control/observation_control_alignment.ipynb
DanWaxman: > **`"same_time"` (the default, and the historical behavior)** pairs $y_k$ with $u_k$:

I'd prefer to not write about historical behavior in tutorials

DanWaxman: Some things are also a bit AI-ey:

> This notebook runs both conventions on the same numbers so the difference is visible rather than asserted.

DanWaxman: is the labeling in the plot for cell 15 correct?

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -89,34 +94,56 @@ def filter_state_dist(state: Any, filter_config: BaseFilterConfig) -> Distributi
...
+        )
+    u = jnp.asarray(u)
+    expected_control_shape = (control_dim,)
+    if u.shape != expected_control_shape:
+        raise ValueError(
+            "control_policy must return one control vector with shape "
+            f"{expected_control_shape}; got {u.shape}."
+        )
+    return u
+
```
DanWaxman: I feel okay with just having this still be in type checking instead of a check

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -89,34 +94,56 @@ def filter_state_dist(state: Any, filter_config: BaseFilterConfig) -> Distributi
...
 
 
+def _validate_policy_control(u: Any, control_dim: int) -> Real[Array, " control_dim"]:
+    """Normalize one control returned by a policy, rejecting bad shapes."""
+    if isinstance(u, Distribution):
+        raise ValueError(
+            "Returning a distribution is not yet supported, instead "
+            "sample from this distribution inside your policy."
+        )
+    u = jnp.asarray(u)
```
DanWaxman: Would avoid this

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -89,34 +94,56 @@ def filter_state_dist(state: Any, filter_config: BaseFilterConfig) -> Distributi
...
-    times -- always passed, even to a policy that ignores them, so that a
-    policy needing genuine time-dependence (e.g. `dynestyx.control.mppi.MPPI`,
-    which plans forward from `t_now`) doesn't need special-casing. Any plain
+    $$u_k, s_{k+1} = \pi(\tilde x_k, t_k, t_{k+1}, s_k)$$
+
+    $\tilde x_k$ is the loop's current state estimate. Which state that is
+    depends on the convention:
+
+    - under `same_time`, $\tilde x_k$ is the predicted state
+    - under `previous_transition`, $\tilde x_k$ is the filtered state
```
DanWaxman: Can you write what the "predicted state" and "filtered state" mean here? Like $x_{k|k}$ etc.

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -89,34 +94,56 @@ def filter_state_dist(state: Any, filter_config: BaseFilterConfig) -> Distributi
...
+
+    It is a NumPyro `Distribution`, depending on the filter configuration.
+    Use `x_hat.mean` for a family-agnostic point estimate.
+    `t_now`/`t_next` are the current and next
+    time points. Any plain
+    callable matching this signature works, including an `equinox.Module`
+    with a matching `__call__`  or a plain Python function.
+
+    `t_now`/`t_next` are the current and next
+    time points. Any plain
```
DanWaxman: already said this

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -151,19 +181,30 @@ class DiscreteControlLoopSimulator(BaseSimulator):
...
+    unspecified (`None`) field resolves to, with a warning.
+
+    With convention `"same_time"`, the control $u_k$ would be chosen *before*
+    seeing the observation $y_k$ it drives, so the policy would see the
+    predicted belief $\tilde p_k$. That needs separate prediction and analysis
+    filter steps, which cuthbert does not expose; an explicit `"same_time"`
+    raises `NotImplementedError` until they are built.
+
+    Both loops are written out in full on the Closed-loop control page
+    (`docs/api_reference/public/control.md`).
```
DanWaxman: Can you make this a link instead?

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -151,19 +181,30 @@ class DiscreteControlLoopSimulator(BaseSimulator):
...
+
+    The one-step filter update currently uses Cuthbert and supports `KFConfig`,
+    `EKFConfig`, `EnKFConfig`, and `PFConfig`. Plated controlled simulation is
+    not yet supported; see
     [Issue #318](https://github.com/BasisResearch/dynestyx/issues/318).
```
DanWaxman: I would prefer things like this defer to links rather than specifics (so that when we add new things we don't forget to update this).

dynestyx/control/discrete_controller_simulators.py
```diff
@@ -271,117 +321,218 @@ def simulate(
...
+                dynamics,
+                rng_key=rng_key,
+                times=times,
+                filter_config=filter_config,
+                initial_policy_state=initial_policy_state,
+            )
+        if alignment == "same_time":
+            raise NotImplementedError(
+                "Closed-loop control with observation_control_alignment='same_time' "
+                "is not implemented yet -- we are working on it. It needs separate "
```
DanWaxman: Can we add an issue to track this?

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -302,6 +302,88 @@ def compute_cuthbert_filter_update(
...
+
+    Maps the filtered belief $\hat p_k$ at `t_prev` to the predicted belief
+    $\tilde p_{k+1}$ at `t`, through the transition driven by `u`:
+
+    $$\tilde p_{k+1}(x) = \int p(x \mid x', u)\, \hat p_k(x')\, dx'.$$
+
+    Not implemented yet. Cuthbert's `Filter` exposes only the fused
+    predict-and-update `filter_combine`, so this step has to be built from
+    cuthbertlib primitives per filter family; that is planned for a
+    follow-up. Closed-loop `"same_time"` control depends on it.
```
DanWaxman: link to issue tracker

DanWaxman: Also, this should be possible for the EnKF (I implemented it as it's necessary for EnRTS filtering)

dynestyx/models/checkers.py
```diff
@@ -235,6 +235,32 @@ def _validate_categorical_state(
...
+        raise ValueError(
+            "observation_control_alignment must be one of "
+            f"{_VALID_OBSERVATION_CONTROL_ALIGNMENTS}, got "
+            f"{observation_control_alignment!r}."
+        )
+    if observation_control_alignment == "previous_transition" and continuous_time:
+        raise ValueError(
+            "observation_control_alignment='previous_transition' is only "
+            "supported for discrete-time models (continuous_time=False)."
+        )
```
DanWaxman: Can we use an enum instead? It makes more obvious what options are available and avoids having to write checkers like this

DanWaxman (changes requested):
This seems reasonable to me overall. I left comments, mostly about docs. But also:
- does it make sense to use an enum instead of a string?
- I think the belief stuff can be implemented for the EnKF already, it gives its predicted states (as it is necessary for the EnRTS algorithm)

docs/tutorials/control/observation_control_alignment.ipynb
MatthieuDarcy: Sorry, this notebook was never supposed to be pushed (I was using it for internal testing). It will be removed.

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Removed

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Removed

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Personally I like it because it gives an explicit error message for when a user gives a bad policy

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Done

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: added a link to the documentation and the relevant files

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: added #372

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: Very nice, where is the implementation?

dynestyx/control/discrete_controller_simulators.py
DanWaxman: Okay, I don't feel super strongly about it (but would remove the `jnp.asarray` call, that should happen way earlier... also lets `u` be typed better in this function)

MatthieuDarcy: yes that part I completely agree with and is removed

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Let me know if you find this sufficiently clear. I'm not as familiar as you are regarding the accepted terminology/notation

dynestyx/inference/integrations/cuthbert/discrete_filter.py
DanWaxman: The states are extracted here: https://github.com/BasisResearch/dynestyx/blob/7c2c93cf7ca43cd3330a76b78c8f09c91ca12e15/dynestyx/inference/observation_predictions.py#L365-L372. The one-step predictive is derived here: https://github.com/BasisResearch/dynestyx/blob/7c2c93cf7ca43cd3330a76b78c8f09c91ca12e15/dynestyx/inference/observation_predictions.py#L299. 

This will require `store_predicted_ensemble=True` to be passed (gets used here:) https://github.com/BasisResearch/dynestyx/blob/7c2c93cf7ca43cd3330a76b78c8f09c91ca12e15/dynestyx/inference/integrations/cuthbert/discrete_filter.py#L305-L316 I think this is currently only done for EnRTS smoother, we should probably have an option in the EnKF config for this. I think this is fine to be a separate PR, sorry, I thought I already added it to the config.

MatthieuDarcy: One (possibly naive) question: why not have this at the cuthbert level? Briefly looking at their code it seems that they have both the `predict` and the`update` step combined in one `filter_combine` step. So it seems that the ingredients are there, just not exposed?

DanWaxman: Their design choice is that `filter_init`/`filter_prepare`/`filter_combine` is more general and they don't want to expose these "intermediary" info unless necessary (e.g., as is necessary in EnRTS). I think @mattlevine22 poked around a bit for this.

dynestyx/models/checkers.py
MatthieuDarcy: We can yes, where would you like it to live? It could be in core where `DynamicalModel` lives

MatthieuDarcy: I put it in core, let me know if you're happy with this

DanWaxman: That seems good, thanks!

dynestyx/control/discrete_controller_simulators.py
DanWaxman: Could you put in an array-like type for `u` then?

dynestyx/control/discrete_controller_simulators.py
DanWaxman: Yup looks good!

DanWaxman (approved):
I made one more reply, commenting about a missing type for control in the validation function. Otherwise, this looks good to me. Thank you Matthieu!!

dynestyx/control/discrete_controller_simulators.py
MatthieuDarcy: Yes, I had to add Distribution to allow for the Distribution check

mattlevine22 (approved):
Looks great, thank you for the hard work on this one!

Issue #327, opened 2026-08-19
Standardize Notation Across Docs


DanWaxman:
We currently have a bit scattered notation. See https://github.com/BasisResearch/dynestyx/pull/307#discussion_r3814511748. We should fix and standardize this

PR #332, opened 2026-08-19
Standardize notation across docs 

MatthieuDarcy:
#327 

Unified some of the API and tutorial notation. 

Choice: 

1. $f$ is used for a generic drift.
2. $L$ is used for a generic diffusion.

This was based on what was the most used convention. 

One additional notation inconsistency: sometimes we use $W_t$, sometimes $\beta_t$ for a Brownian motion. Might be worth just using $W_t$ throughout.


DanWaxman (approved):
Thanks! I found a few more $\mu$s and a $g$ and fixed those, plus got rid of the few uses of $\beta_t$ I saw around. Looks good to me now, will merge after tests are green!

PR #333, opened 2026-08-19
Add cuthbert EnKF support for observation scoring

DanWaxman:
Currently, observation scoring is only available in continuous-time, as the CD-Dynamax filters generally return a canonical predictive solution to the Chapman-Kolmogorov eq. and cuthbert does not. EnKF is a notable exception; v0.0.14 of cuthbert introduced saving the ensemble, as this is required for downstream computation of EnRTS. This adds support for the observation scoring pattern to EnKF, and provides a notebook applying it to ODEs.


dynestyx/inference/observation_predictions.py
```diff
@@ -46,6 +51,38 @@ class PredictedObservationOutputs:
     noise_cov: Float[Array, "*plate time observation_dim observation_dim"] | None = None
 
 
+def _gaussian_observation_mean_and_covariance(
```
mattlevine22: this looks super general, which is nice---might be worth removing "observation" from this

dynestyx/inference/observation_predictions.py
```diff
@@ -286,9 +432,44 @@ def extract_continuous_filter_predictions(
...
+            posterior,
+            dynamics=dynamics,
+            plate_shapes=plate_shapes,
+        )
 
-    return _build_prediction_outputs(
+    return None
+
+
+def extract_continuous_filter_predictions(
```
mattlevine22: thin function used only once?

mattlevine22 (changes requested):
Two small comments, but looks great

mattlevine22 (approved):
LGTM

Issue #330, opened 2026-08-19
Fix tutorial notebooks


MatthieuDarcy:
`heat_equation_inference.ipynb` is now borken.

DanWaxman:
Closed by #334.

PR #334, opened 2026-08-19
Fix heat_equation_inference notebook

MatthieuDarcy:
#330 fix. I also coarsened the discretization to make this run faster


DanWaxman:
Thanks!! To be clear: is it just that the `ODESimulator` no longer supports inference, or was the reparameterization also required to make inference work now?

MatthieuDarcy:
Both. 
I had to update to use `LatentPathBuilder` (and fix some import issues).

The reparametrization was always required but now had to be done diffrently (by explicitly building `u_0 = initial_condition(mu) + ic_noise * z`), it seems that now `TransformReparam` doesn't work?

DanWaxman:
Ah interesting... I'll look at the reparam thing later, I'd prefer we are able to use that.

DanWaxman:
Okay this turned out to be because `LatentPathBuilder` gets rid of a dedicated `numpyro` site for the initial condition that can be reparameterized. I've opened #341 to deal with this. I'll approve this in the meantime, though it would be great to merge them together as a stack if you and @mattlevine22 don't have qualms with the fix.

DanWaxman (approved):

PR #339, opened 2026-08-20
Use EnKF on ODE using Discretizer

DanWaxman:
This uses the EnKF on the ODE using the new ODEFlowConfig for Discretizer, which let's us avoid adding spurious noise. This requires a small fix to discrete_filter, where we avoid casting an array to float. I also narrowed the search range for $\delta$ to $[0.0, 0.10)$ since it looks a bit nicer.


mattlevine22 (approved):
Looks great. FYI at the end the text says `delta = 0.1` for naive, but `0.5` is actually used.

PR #341, opened 2026-08-24
Repair compatibility with NumPyro reparam machinery

DanWaxman:
After moving to `LatentPathBuilder`, numpyro's reparam machinery no longer worked. This is because there was no longer a dedicated numpyro site corresponding to initial conditions. This is now patched directly, and implemented in the heat equation tutorial notebook.


mattlevine22 (approved):
quite reasonable

PR #343, opened 2026-08-27
Add EnKF Localization

DanWaxman:
This PR adds EnKF localization to `dynestyx`, following its recent addition to `cuthbert`. This allows for two modes of using localization: localization via a taper function and pairwise distances, and using callbacks that modify the covariance directly. These mirror the cuthbert implementations. 

We also include a new tutorial notebook that shows how localization can be applied, and a deep dive notebook that shows how localization hyperparameters can be learned via gradient descent on a proper scoring rule.


docs/tutorials/gentle_intro/13_enkf_localization.ipynb
mattlevine22, line 1: - Overall, I'd like this tutorial's narrative to be a bit more clear that it is ultimately building L_xy. It should show the different options you've built, including a more "automatic" way and a more "manual" way (so you can walk away from the tutorial with copy-paste stuff to do any kind of localization you want). I think we should be thorough about the code here; people reading this will know the math pretty well, so it is really a code tutorial in my view (it is still important to keep the math accessible and well motivated, etc.). I see now that the custom stuff is at the bottom---I might suggest flipping the order. First, say, here is the generic way to build an L_xy and explain what `state_observation_distances` is (this can be a bit confusing). Then say: we've implemented some popular ones, like Gaspari-Cohn...etc. 
---> TLDR; I think we should focus on "how to do any localization" in this tutorial rather than lead towards a best practice choice.

Details:

- The definition of L96 equations need to be a bit more specific about defining the maximum coordinate K and how things wrap around the 0 boundary, etc.
- I also think the way state-noise appears in the system (and observations) should be made explicit. Let's use Andrew's standard notation...x_k+1 = Psi(x_k) + nu_k; y_k = Hx_k + eta_k (or other greek letters / u / v etc.), with Psi the Tsit solver.
- in the results, I'm a bit confused why the Unlocalized filter in red starts out "synchronized" when supposedly you've thrown out 50 cycles? 1) "cycles" should be defined more clearly (probably just T), 2) the burn-in story should be made more explicit...currently, it looks like unlocalized-filter happened to desynchronize around "T=50 cycles", which is rather odd, 3) might help to plot something that shows the full simulated data and what was thrown out.

- instead of using `initial_mean = truth[0] + 0.5 * jr.normal(jr.PRNGKey(2027), (STATE_DIM,))`, it is common to use samples from the invariant measure. Just a small thing, definitely don't have to do it.

- Building periodic distance matrix: I worry this is too specific and confusing for a gentle intro (i.e. periodic stuff). Can it intro the common stories? a) if your indices are arbitrary, be careful,  b) if your indices represent a fixed spatial grid, do this; ..., c) if your indices represent a periodic spatial grid, like this, do this;

docs/deep_dives/l96_localization_hyperparameter_scoring.ipynb
mattlevine22, line 1: - again, more clearly define L96 with indices.
- again, I wouldn't use "time indices 50–199", I'd specify things in model time units and define the observation sample rate. Later it says "time units" ambiguously, and also has N cycles in the code :)
- "produces non-PSD covariance matrices when using the Gaussian taper instead of the more common Gaspari-Cohn taper"...maybe a disclaimer here that you're using the Gaussian Taper anyway for demo purposes?
- "positive-definite Gaussian taper on the periodic domain" I think this needs written math to go with it.

- Can we search a wider space than `LENGTH_SCALE_GRID = jnp.linspace(0.5, 6.0, 24)`, since 6.0 is optimal? Or do things misbehave above 6? And why bound LBFGS at 6 too?

- with the holdout stuff, I'm not sure this is really a fair comparison. What actually happens is the filters loses synchrony and can never come back (probably due to collapse). at the very least, we should take the unseen parts of the trajectory and filter using an un-informed initial condition (rather than informed by training data). Also, you may like using i.c. as sampled from invariant measure instead of normal from around the truth (it is a cool trick that readers may like, but not a big deal)

docs/api_reference/public/inference/configs/filter_configs.md
```diff
@@ -20,6 +20,61 @@ The single `Filter()` handler is directed to the appropriate filtering algorithm
 | `ContinuousTimeEnKFConfig` | Continuous-discrete | High-dimensional or expensive models with lower-dimensional structure and Gaussian observations (approximate). Performs reasonably as a default. *(default)* |
 | `ContinuousTimeDPFConfig`  | Continuous-discrete | Applicable for arbitrary state-space models, but quite expensive and noisy estimates (asymptotically exact in the limit of infinite particles, approximate in practice). |
 
+## EnKF localization
+
+Localization is a common technique in EnKFs, related to the issue of estimating a covariance matrix from a few samples. The (cuthbert, discrete-time) EnKF supports localization via covariance tapering via pairwise distances, or direction control over localization functions, using `EnKFLocalizationConfig` for the former and `EnKFLocalizationFunctions` for the latter. For localization in a deterministic continuous-time dynamics, combine `EnKFConfig` with an ODE-flow `Discretizer`.
```
mattlevine22: If someone wants direct control, where would you point them in our documentation / tutorials?

dynestyx/inference/configs/filter.py
```diff
@@ -23,6 +26,153 @@
...
+        self,
+        predicted_observation_covariance: Array,
+        model_inputs: Any,
+    ) -> Array:
+        """Return a predictive covariance with the same shape as the input."""
+        ...
+
+
+@dataclasses.dataclass
+class EnKFLocalizationConfig:
```
mattlevine22, line 70: Maybe we should have a deep dive that shows how to build your own stuff? Maybe even re-implement the existing features and run them both side-by-side and show we get the same answer?

pyproject.toml
```diff
@@ -31,9 +31,9 @@ classifiers = [
 ]
 dependencies = [
     "effectful>=0.4.0",
-    "cuthbert==0.0.14",
-    "cuthbertlib==0.0.14",
-    "cd-dynamax>=0.4.3",
+    "cuthbert==0.0.15",
+    "cuthbertlib==0.0.15",
+    "cd-dynamax>=0.4.1",
```
mattlevine22: why downgrade cd-dynamax?

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -346,27 +383,54 @@ def compute_cuthbert_filter(
...
+        )
+
+    if (
+        resolved_localization is not None
+        and resolved_localization.observation_taper is not None
+    ):
+        observation_taper = resolved_localization.observation_taper
+        cuthbert_inputs = _LocalizedCuthbertInputs(
+            **input_kwargs,
+            localization_observation_taper=jnp.broadcast_to(
```
mattlevine22: Codex says:

* Disabling predictions still stores a quadratic-size matrix at every timestep. This unconditionally broadcasts the observation taper across time and retains it in the returned states. I confirmed this with include_predicted_observations=False. For 1,000 observations over 1,000 timesteps, this adds roughly 4 GB in float32 just for repeated taper values. Only retain this scoring metadata when needed, preferably storing the constant matrix once.

mattlevine22 (changes requested):
Overall this looks great.

Most of my feedback really revolves around finding simple + clear ways to show a new user (e.g., who already knows DA + localization) how to do what they want very clearly.

I think of all my feedback, the most salient point may be to add a deep dive where you demo the custom localization interface and re-implement gaspari-cohn, etc. and show that you get the same results as the automatic interface. And then make sure we link that deep dive at the top of the gentle intro, etc.

pyproject.toml
DanWaxman: Oops, mistake when resolving merge conflicts. Thanks!

dynestyx/inference/integrations/cuthbert/discrete_filter.py
DanWaxman: This should be fixed now, we resolve up-front and use that resolved input across all time steps.

Issue #337, opened 2026-08-20
Feature improvements for control


MatthieuDarcy:
Here are some improvements that would benefit our control implementation

1. Add controls as `SimulatedResult` when running open loop control (useful for example in MPC). 
2. Improve MPPI proposal distribution: right now it uses white noise , but AR(1) and Colored noise would be significant improvements.

Related but more significant
#336 and #312.

PR #348, opened 2026-08-31
Fixed MPPI

MatthieuDarcy:
Solves part of #337 .

MPPI now uses the `previous_transition` convention from `dsx.simulate` directly.

The implementation is now mathematically  correct: it does not duplicate controls.

The initial state $x_0$ or the initial observation $y_0$ are not present in the `SimulatedResult` fields`states` and `observations`, which will avoid a user using them in their loss accidentally. The initial condition $x_0$ is available through `res.x_0`. This is intuitive from a user perspective (you basically never want to use the initial condition in your loss), but differs from the usual contract, I don't know if its a good idea.


MatthieuDarcy:
Now uses `vmap` directly over `SimulatedResult` thanks to #350

dynestyx/control/mppi.py
```diff
@@ -118,36 +124,22 @@ def _rollout_and_score_one(
...
-        # (below), so pin that convention here whatever the model says: the
-        # outer closed loop may run the same model as "previous_transition".
-        # is_leaf lets tree_at replace the field when it is None (unspecified).
+        # Pin the rollout to start at x0, and plan under the  "previous_transition"
+        # convention so y_{k+1} and x_{k+1} both produced by u_k.
         pinned_dynamics = eqx.tree_at(
             lambda m: (m.initial_condition, m.observation_control_alignment),
             self.dynamics,
-            (dist.Delta(x0, event_dim=1), "same_time"),
+            (dist.Delta(x0, event_dim=1), "previous_transition"),
```
mattlevine22: This silent override might confuse a user---is there a check + warning of the override that happens earlier on?

tests/test_discrete_control.py
```diff
@@ -1370,3 +1370,106 @@ def flaky_loss(result):
...
+    )
+
+    nominal = jnp.array([[1.0], [2.0], [3.0]])
+    _, _, result = mppi.plan_step(
+        dist.Delta(x_0).to_event(1),
+        t_now,
+        (nominal, jr.PRNGKey(0)),
+    )
+
+    # Plain SimulatedResult now carries the controls; no ControlledSimulatedResult.
```
mattlevine22: unnecessary

mattlevine22 (changes requested):

tests/test_discrete_control.py
MatthieuDarcy: removed

dynestyx/control/mppi.py
MatthieuDarcy: The override is in the description of MPPI itself (as far as I know its the most natural definition of the MPPI algorithm), but that's about it. The override is also purely internal to MPPI, it doesn't modify the dynamics otherwise.

MatthieuDarcy: I think we could add it in the class definition with default previous transition (matching the closed loop control default).

mattlevine22: well the point is that you can hand it dynamics with "same_time" and not realize it is going to overwrite that. 

Maybe add a `__post_init__` check that warns of the override if a user has specified "same_time"?

MatthieuDarcy: yes that's easy, done

mattlevine22 (approved):
Looks good!

PR #350, opened 2026-09-01
Make SimulatedResult an Equinox module

### LINKED ISSUE #347: Support batched control sequences in `dsx.simulate`
`dsx.simulate(..., n_simulations=N)` currently draws `N` trajectories that differ only
in their initial condition. 

Sampling based MPC (such as MPPI) would like to batch over control sequences. 

### Tentative recommendation 

Allow `ctrl_values` to carry an optional leading simulation axis:

```python
ctrl_values: Real[Array, "ctrl_time control_dim"]          # shared (current behavior)
           | Real[Array, "n_controls ctrl_time control_dim"]  # per-trajectory
```
We would have to think about how this interact with `n_simulations`.
1. Option 1: creates $N \times n_c$ simulation (each control has $N$ simulations attached). Could be useful if we want to have the same $N$ initial conditions/randomness, each paired with a seperate control. This is the most useful in MPPI.
2. Option 2: assume that $N= n_c$, sample initial conditions/randomness seperately per control. Note that a user could duplicate controls to achieve $N\times n_c$ trajectories (with the caveat that randomness is not shared in this case).

DanWaxman:
Resolves #347 by creating as an equinox module.

### DIFF
--- a/dynestyx/control/discrete_controller_simulators.py
+++ b/dynestyx/control/discrete_controller_simulators.py
@@ -1,6 +1,5 @@
 """Closed-loop simulation for controlled discrete-time dynamical models."""
 
-import dataclasses
 from typing import Any, Protocol, runtime_checkable
 
 import jax
@@ -102,7 +101,6 @@ class PolicyCallable(Protocol):
         raise NotImplementedError()
 
 
-@dataclasses.dataclass
 class ControlledSimulatedResult(SimulatedResult):
     """`SimulatedResult` extended with the control loop's extra outputs.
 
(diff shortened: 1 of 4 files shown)


mattlevine22 (approved):
Thanks Dan, this looks solid to me!

MatthieuDarcy (commented):
Thanks Dan, this seems like a good solution (my Claude is very enthusiastic lol).

DanWaxman:
> Thanks Dan, this seems like a good solution (my Claude is very enthusiastic lol).

I'm glad I can still outsmart and elate our robot overlords (for now) 😆

PR #352, opened 2026-09-01
Make result containers JAX-compatible Equinox modules

mattlevine22:
Converts array-bearing result containers to Equinox modules for JAX/vmap compatibility while preserving immutable handler composition. 

Adds focused vmap coverage for result stacking and `dsx.condition`.


DanWaxman:
As discussed, I think this is not doable for now, for the reason that numpyro distributions carry lots of static metadata and don't always vmap the way you'd expect. Will close for now, we can re-open after figuring out the dist stuff.

PR #354, opened 2026-09-02
Vectorize summarize_gaussian_mixture to avoid slow jax concatenate

DanWaxman:
We had a big `jnp.stack` in the `summarize_gaussian_mixture` function of the missingness tutorial notebook. There is ostensibly a regression in stacking on minor axes in `jax >= 0.11.0` that can cause bad runtime here. This is solved by vmapping directly instead of forming a list and stacking.


mattlevine22 (approved):
LGTM

Issue #355, opened 2026-09-03
Use `dist.LowRankMultivariateNormal` for EnKF


MatthieuDarcy:
Right now all Gaussian based filtering algorithms use a dense covariance matrix of size $d \times d$ formed from cholesky factors. The EnkF has a low rank structure (rank $N-1$) which can be expressed using `dist.LowRankMultivariateNormal`, which reduces the complexity to $N\ times d$, making it much more efficient for large $d$.

PR #356, opened 2026-09-03
Changed the EnKF distribution to be low rank

MatthieuDarcy:
# Summary
Adresses #355. 

Return LowRankMultivariateNormal for rank-deficient EnKF distributions (for both `Filter` and `Smooth`)

Previously the distributions returned for the EnKF were always `MultivariateNormal`, with the covariance formed densely from the QR factors. When the ensemble covariance was low rank (`n_particles - 1 < state_dim`) this had two consequences:

1. The returned distributions were singular and yielded NaN when `.sample` or `.log_prob` were used. NumPyro accepts a singular covariance matrix, which then gives NaN downstream when it is factorized.

2. It was computationally expensive for large state_dim.

The solution is to check whether the covariance is singular (a simple comparison of `n_particles` against `state_dim`) and return a `LowRankMultivariateNormal` in that case, which can be sampled. The same is done in filter_state_dist, used by DiscreteControlLoopSimulator. If in the non degenerate case, we use the same `MultivariateNormal` to ensure existence of densities.

Note that this affects only the returned distributions. The filtered/smoothed states are unchanged and behavior is otherwise the same as before (`marginal_loglik` is identical).

One design choice I am unsure of: I added a jitter parameter to the EnKF config (filtered_covariance_jitter, default 0.0). LowRankMultivariateNormal has no valid log_prob unless a jitter is added, so this lets the user obtain a distribution with a valid density when needed. The jitter does not affect the ability to sample, but it does change the resulting distribution.

# Bug fixes
Claude claims to have solved 2 bugs:

1. In `DiscreteControlLoopSimulator`, filter_state_dist passed the rectangular ensemble factor as scale_tril, so the belief reported event_shape == (n_particles,). `.mean` and `.log_prob` would then raise an error.
2. Posterior rollout (predict_times with a filtered_result) returned all-NaN states: it grafts the filtered distributions in as the forecast initial condition and samples them, and those samples were NaN because the dense covariance was singular.

# Verification 

The test suite now ensures the correct behavior.

I'm still double checking that this yields the same results as before but I locally ran the following tests:

1. Equivalence on a single distribution for the same ensemble at various dimensionality and ensemble size ($d= 8, 16, 64, 256, N = 8, 16, 32$).
2. End-to-end filtering: same `marginal_loglik` as before, distributions agree when both work, new low rank works when the distributioj is degenerate.
3. Speed test: $d = 1024, 4096, 10 000$, measures the speedup and memory saving (between 7x-608x improvement).

### DIFF
--- a/dynestyx/inference/configs/discretizer.py
+++ b/dynestyx/inference/configs/discretizer.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import abc
 import dataclasses
-import math
 
 import diffrax as dfx
 
@@ -12,21 +11,15 @@ from dynestyx.inference.configs.simulator import (
     ODESimulatorConfig,
     SDESimulatorConfig,
 )
+from dynestyx.utils import _validate_nonnegative_float
 
 
 def _validate_covariance_jitter(covariance_jitter: float) -> None:
-    if not math.isfinite(covariance_jitter) or covariance_jitter < 0.0:
-        raise ValueError(
-            "covariance_jitter must be a finite, nonnegative float, "
-            f"got {covariance_jitter!r}."
-        )
+    _validate_nonnegative_float("covariance_jitter", covariance_jitter)
 
 
 def _validate_jitter_scale(jitter_scale: float) -> None:
-    if not math.isfinite(jitter_scale) or jitter_scale < 0.0:
-        raise ValueError(
-            f"jitter_scale must be a finite, nonnegative float, got {jitter_scale!r}."
-        )
+    _validate_nonnegative_float("jitter_scale", jitter_scale)
 
 
 def _default_diffrax_sde_solver() -> SDESimulatorConfig:
--- a/dynestyx/inference/configs/filter.py
+++ b/dynestyx/inference/configs/filter.py
@@ -22,6 +22,17 @@ FilterSource = (
 CuthbertOrCDDynamaxFilterSource = CuthbertOnlyFilterSource | CDDynamaxOnlyFilterSource
 
 
+def _validate_filtered_covariance_jitter(filtered_covariance_jitter: float) -> None:
+    if (
+        not math.isfinite(filtered_covariance_jitter)
+        or filtered_covariance_jitter < 0.0
+    ):
+        raise ValueError(
+            "filtered_covariance_jitter must be a finite, nonnegative float, "
+            f"got {filtered_covariance_jitter!r}."
+        )
+
+
 @dataclasses.dataclass
 class BaseFilterConfig(abc.ABC):
     r"""Shared configuration options inherited by all filter configs.
@@ -134,6 +145,11 @@ class EnKFConfig(BaseFilterConfig):
         inflation_delta (float | None): Scale ensemble anomalies by
             \(\sqrt{1 + \delta}\) before the update to prevent collapse.
             `None` disables inflation.
+        filtered_covariance_jitter (float): Nonnegative \(\epsilon\) added to the
+            reported filtered-state covariance as \(\epsilon I\). Defaults to `0.0`.
+            When the filtered covariance is singular (i.e., when the ensemble size is smaller than the state dimension),
+            this jitter ensures that the resulting distribution has a well-defined density.
+            It never affects the filter recursion or the marginal likelihood.
         filter_source (FilterSource): Backend. Defaults to `"cuthbert"`.
 
     ??? note "Algorithm Reference"
@@ -185,7 +201,11 @@ class EnKFConfig(BaseFilterConfig):
     )
     perturb_measurements: bool | None = None
     inflation_delta: float | None = None
+    filtered_covariance_jitter: float = 0.0
     filter_source: CuthbertOnlyFilterSource = "cuthbert"
+
+    def __post_init__(self) -> None:
+        _validate_filtered_covariance_jitter(self.filtered_covariance_jitter)
 
 
 @dataclasses.dataclass
--- a/dynestyx/inference/configs/filter.py
+++ b/dynestyx/inference/configs/filter.py
@@ -8,6 +8,10 @@ from typing import Literal
 import jax.random as jr
 from jaxtyping import PRNGKeyArray
 
+from dynestyx.inference.utils.distribution_utils import (
+    CovarianceJitter,
+    _default_covariance_jitter,
+)
 from dynestyx.utils import _validate_nonnegative_float
 
 ResamplingBaseMethod = Literal["systematic", "multinomial", "stratified"]
@@ -136,11 +140,17 @@ class EnKFConfig(BaseFilterConfig):
         inflation_delta (float | None): Scale ensemble anomalies by
             \(\sqrt{1 + \delta}\) before the update to prevent collapse.
             `None` disables inflation.
-        filtered_covariance_jitter (float): Nonnegative \(\epsilon\) added to the
-            reported filtered-state covariance as \(\epsilon I\). Defaults to `0.0`.
-            When the filtered covariance is singular (i.e., when the ensemble size is smaller than the state dimension),
-            this jitter ensures that the resulting distribution has a well-defined density.
-            It never affects the filter recursion or the marginal likelihood.
+        recorded_filtered_states_cov_jitter (float | Literal["auto"]): Nonnegative \(\epsilon\) added to
+            the **recorded** filtered-state covariance as \(\epsilon I\).
+            This only affects the covariance when converted to a `MultivariateNormal` or `LowRankMultivariateNormal`
+            distribution (notably those returned in `ConditionedResult.dists`); it never
+            enters the EnKF update, the filter recursion, or the marginal
+            likelihood.
+            When `n_particles - 1 < state_dim`, the ensemble covariance is singular,
+            this regularization is necessary to give the recorded distributions a well-defined density (sampling will work nonethelss).
+            `"auto"` (default) selects a small precision-dependent value
+            (`1e-5` in float32, `1e-12` in float64). Will work for variance around 1, but may need a bigger value
+            for larger magnitudes. Pass `0.0` for the exact, unregularised covariance.
         filter_source (FilterSource): Backend. Defaults to `"cuthbert"`.
 
     ??? note "Algorithm Reference"
@@ -192,12 +202,16 @@ class EnKFConfig(BaseFilterConfig):
     )
     perturb_measurements: bool | None = None
     inflation_delta: float | None = None
-    filtered_covariance_jitter: float = 0.0
+    recorded_filtered_states_cov_jitter: CovarianceJitter = "auto"
     filter_source: CuthbertOnlyFilterSource = "cuthbert"
 
     def __post_init__(self) -> None:
+        if self.recorded_filtered_states_cov_jitter == "auto":
+            self.recorded_filtered_states_cov_jitter = _default_covariance_jitter()
+        # Check that the jitter is nonnegative float
         _validate_nonnegative_float(
-            "filtered_covariance_jitter", self.filtered_covariance_jitter
+            "recorded_filtered_states_cov_jitter",
+            self.recorded_filtered_states_cov_jitter,
         )
 
 
--- a/dynestyx/inference/configs/filter.py
+++ b/dynestyx/inference/configs/filter.py
@@ -8,6 +8,8 @@ from typing import Literal
 import jax.random as jr
 from jaxtyping import PRNGKeyArray
 
+from dynestyx.utils import _validate_nonnegative_float
+
 ResamplingBaseMethod = Literal["systematic", "multinomial", "stratified"]
 ResamplingDifferentiableMethod = Literal["stop_gradient", "straight_through", "soft"]
 FilterEmissionOrder = Literal["zeroth", "first", "second"]
@@ -20,17 +22,6 @@ FilterSource = (
     CuthbertOnlyFilterSource | CDDynamaxOnlyFilterSource | DynestyxOnlyFilterSource
 )
 CuthbertOrCDDynamaxFilterSource = CuthbertOnlyFilterSource | CDDynamaxOnlyFilterSource
-
-
-def _validate_filtered_covariance_jitter(filtered_covariance_jitter: float) -> None:
-    if (
-        not math.isfinite(filtered_covariance_jitter)
-        or filtered_covariance_jitter < 0.0
-    ):
-        raise ValueError(
-            "filtered_covariance_jitter must be a finite, nonnegative float, "
-            f"got {filtered_covariance_jitter!r}."
-        )
 
 
 @dataclasses.dataclass
@@ -205,7 +196,9 @@ class EnKFConfig(BaseFilterConfig):
     filter_source: CuthbertOnlyFilterSource = "cuthbert"
 
     def __post_init__(self) -> None:
-        _validate_filtered_covariance_jitter(self.filtered_covariance_jitter)
+        _validate_nonnegative_float(
+            "filtered_covariance_jitter", self.filtered_covariance_jitter
+        )
 
 
 @dataclasses.dataclass
--- a/dynestyx/inference/filters.py
+++ b/dynestyx/inference/filters.py
@@ -716,7 +716,9 @@ class Filter(BaseLogFactorAdder):
                 states,
                 particle_mode=isinstance(config, PFConfig),
                 plate_shapes=plate_shapes,
-                covariance_jitter=getattr(config, "filtered_covariance_jitter", 0.0),
+                covariance_jitter=getattr(
+                    config, "recorded_filtered_states_cov_jitter", 0.0
+                ),
             )
 
         raise ValueError(f"Unsupported batched output kind: {output_kind}")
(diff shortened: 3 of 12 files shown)


dynestyx/inference/configs/filter.py
```diff
@@ -22,6 +22,17 @@
 CuthbertOrCDDynamaxFilterSource = CuthbertOnlyFilterSource | CDDynamaxOnlyFilterSource
 
 
+def _validate_filtered_covariance_jitter(filtered_covariance_jitter: float) -> None:
```
mattlevine22: `inference/configs/discretizer.py` already has `_validate_jitter_scale` and `_validate_covariance_jitter`.

Can we use those?

dynestyx/inference/configs/filter.py
```diff
@@ -134,6 +145,11 @@ class EnKFConfig(BaseFilterConfig):
         inflation_delta (float | None): Scale ensemble anomalies by
             \(\sqrt{1 + \delta}\) before the update to prevent collapse.
             `None` disables inflation.
+        filtered_covariance_jitter (float): Nonnegative \(\epsilon\) added to the
```
mattlevine22: I'd worry a user thinks this will appear in the EnKF algorithm (which it doesn't).

How about a name like `recorded_filtered_states_[lowrank]_cov_jitter`?


Do we want this jitter usable even for full rank MVN? or just to deal w LowRank stuff

dynestyx/inference/configs/filter.py
```diff
@@ -185,8 +201,12 @@ class EnKFConfig(BaseFilterConfig):
     )
     perturb_measurements: bool | None = None
     inflation_delta: float | None = None
+    filtered_covariance_jitter: float = 0.0
```
mattlevine22: Maybe `recorded_filtered_states_cov_jitter`?

dynestyx/inference/utils/distribution_utils.py
```diff
@@ -174,18 +244,42 @@ def _cholesky_state_sequence_to_dists(
...
 
+    if hasattr(states, "ensemble") and _check_if_ensemble_low_rank(states.ensemble):
+        return _ensemble_sequence_to_low_rank_gaussian_dists(
+            states.ensemble,
+            covariance_jitter=covariance_jitter,
+            plate_shapes=plate_shapes,
+        )
+
+    covariances = covariance_from_cholesky(states.chol_cov)
+    if covariance_jitter:
```
mattlevine22, line 277: so this uses the jitter in the full-rank MVN...do we want to even allow that?

tests/test_filters.py
```diff
@@ -669,6 +669,41 @@ def test_cuthbert_enkf_sparse_h_matches_dense_h():
...
+    with Filter(
+        filter_config=EnKFConfig(n_particles=n_particles, crn_seed=jr.PRNGKey(42))
+    ):
+        result = dsx.condition(
+            "f", dynamics, obs_times=obs_times, obs_values=obs_values
+        )
+
+    ensemble = result.states.ensemble
+    assert len(result.dists) == len(obs_times)
+    for t, d in enumerate(result.dists):
```
mattlevine22, line 727: Since this test uses default `...jitter=0.0`, a `d.log_prob(...)` call would fail / NaN here right?

May want a regression test specifically that jitter protects you in that case (its the failure mode that made you build it).

mattlevine22 (commented):
looks quite reasonable to me! 

Question: Do we want the new jitter to be used in regular MVN stuff too (currently it is)?

My recommendation would be:
- only use this jitter for the low rank MVN
- give it a > 0 default (so it protects from the .log_prob nan issue)
- rename / redoc this in the EnKF config to be clear that it is only for low rank and only for the recorded dists (I suggested a name above)

Certainly open to the current choice if you think its worth it.

MatthieuDarcy:
> looks quite reasonable to me!
> 
> Question: Do we want the new jitter to be used in regular MVN stuff too (currently it is)?
> 
> My recommendation would be:
> 
>     * only use this jitter for the low rank MVN
> 
>     * give it a > 0 default (so it protects from the .log_prob nan issue)
> 
>     * rename / redoc this in the EnKF config to be clear that it is only for low rank and only for the recorded dists (I suggested a name above)
> 
> 
> Certainly open to the current choice if you think its worth it.

Hard agree on the "making it clear that this only affects the recorded dists" point, I'll clarify this.

Regarding its use in MVN: I think there is an argument to allow its use in MVN because this case can also be singular for other reasons and this yields a single keyword to regularize the returned dists in all cases (the internal distinction is not very intuitive for a user, no?)

Regarding its default value, I support  giving it a > 0 default. This is technically lying to the user because the posterior distributions are singular, but this singular behavior comes from the numerical method (and not some intrisic property of the system) so regularizing is a likewise a valid numerical choice. It does not affect MCMC, but changes things like scoring from the filtered distribution (which is relevant when doing a forecast). 

My tentative recommendation is:

1. Allow its use for MVN. 
2. Give it a default > 0 (possibly set it = 0 for the MVN case, if we want to leave the MVN case untouched).

Open to changing 1 especially if there are other types of safeguard against singular behavior in this case.

mattlevine22:
I support your recommendation---and I think a default of something small is fine to use across the board.

MatthieuDarcy:
# Summary

Renamed `filtered_covariance_jitter` to `recorded_filtered_states_cov_jitter`.

Adopted convention: `EnKF` hasa positive default via an `"auto"` sentinel:  either `1e-5` or `1e-12` depending on resolves once in `EnKFConfig.__post_init__` to a precision-dependent value (1e-5 in
float32, 1e-12 in float64). Compute so that variance ~ 1 should work, but may fail for larger variances. This is true no matter the low rank/full rank distinction, see #357 for a consideration for why we might want to discriminate between the two.

In control, the old `filter_state_dist` now delegates to `_cholesky_state_sequence_to_dists` with the same arguments `run_discrete_filter` passes it, instead of duplicating the three-branch dispatch, thus unifying with the behavior from filter.

_validate_jitter_scale and _validate_covariance_jitter were redundant. Created a new utility `_validate_nonnegative_float` into `dynestyx/utils.py` which all jitter validators can now call. Each caller passes its own field name to ensure so error messages still name the attribute the user set.

Simplified and improved test suite: 
`test_ensemble_jitter_is_off_by_default_and_enabled_only_by_config` verifies that each filter config has the right jitter value and that manually setting a jitter value gives the right answer.
`test_covariance_jitter_shifts_only_the_covariance_diagonal` checks the perturbation is exactly `eps * I` and the mean untouched, on both the dense and low-rank branches.
`test_cuthbert_enkf_filtered_dists_are_low_rank_and_samplable` rebuilds the dists as `run_discrete_filter` does and asserts they are no longer singular.

dynestyx/inference/configs/filter.py
MatthieuDarcy: These were doing the same thing, but with different names in the error message. I added a `_validate_nonnegative_float` in `dynestyx/utils.py`that can be used more generally (it takes in a name so that the user knows which parameter is wrong in case of error).

dynestyx/inference/configs/filter.py
MatthieuDarcy: Changed this per our discussion.

tests/test_filters.py
MatthieuDarcy: I improved the tests. There are now 3 tests: first a check that the filter configs default to the correct values, second that the returned distributions are correct (low rank, jitter only perturbs the diagonal...) and a seperate test with default EnKF to make sure that the density is well defined.

dynestyx/inference/utils/distribution_utils.py
```diff
@@ -120,6 +124,26 @@ def _gaussian_sequence_to_dists(
     ]
 
 
+_DEFAULT_COV_JITTER_F32 = 1e-5
+_DEFAULT_COV_JITTER_F64 = 1e-12
+
+
+def _default_covariance_jitter() -> float:
```
mattlevine22: @DanWaxman is this dangerous?

dynestyx/inference/integrations/cuthbert/discrete_filter.py
```diff
@@ -436,7 +436,9 @@ def run_discrete_filter(
     filtered_dists = _cholesky_state_sequence_to_dists(
         states,
         particle_mode=isinstance(filter_config, PFConfig),
-        covariance_jitter=getattr(filter_config, "filtered_covariance_jitter", 0.0),
+        covariance_jitter=getattr(
+            filter_config, "recorded_filtered_states_cov_jitter", 0.0
```
mattlevine22, line 440: do we need/want 0.0 if it has an auto mode /default  now?

dynestyx/inference/integrations/cuthbert/discrete_smoother.py
```diff
@@ -289,7 +289,9 @@ def run_discrete_smoother(
     smoothed_dists = _cholesky_state_sequence_to_dists(
         states,
         particle_mode=isinstance(smoother_config, PFSmootherConfig),
-        covariance_jitter=getattr(smoother_config, "filtered_covariance_jitter", 0.0),
+        covariance_jitter=getattr(
+            smoother_config, "recorded_filtered_states_cov_jitter", 0.0
```
mattlevine22, line 293: do we need/want 0.0 if it has an auto mode /default now?

dynestyx/inference/filters.py
```diff
@@ -716,7 +716,9 @@ def compute_output_member(dyn, ot, ov, ovf, om, ct, cv, k, *idxs):
                 states,
                 particle_mode=isinstance(config, PFConfig),
                 plate_shapes=plate_shapes,
-                covariance_jitter=getattr(config, "filtered_covariance_jitter", 0.0),
+                covariance_jitter=getattr(
+                    config, "recorded_filtered_states_cov_jitter", 0.0
```
mattlevine22, line 720: do we need/want 0.0 if it has an auto mode /default now?

dynestyx/inference/smoothers.py
```diff
@@ -582,7 +582,9 @@ def compute_output_member(dyn, ot, ov, ct, cv, k, *idxs):
                 states,
                 particle_mode=isinstance(config, PFSmootherConfig),
                 plate_shapes=plate_shapes,
-                covariance_jitter=getattr(config, "filtered_covariance_jitter", 0.0),
+                covariance_jitter=getattr(
+                    config, "recorded_filtered_states_cov_jitter", 0.0
```
mattlevine22, line 586: do we need/want 0.0 if it has an auto mode /default now?

dynestyx/inference/integrations/cuthbert/discrete_filter.py
MatthieuDarcy: In the current implementation yes. The idea is that all other filters (KF etc...) do not have a jitter attribute, this sets it to 0.0 for those filters.

dynestyx/inference/utils/distribution_utils.py
DanWaxman: I think I'm okay with it here... though in principal we should be determining this based on the actual data dtype instead of the default dtype.

MatthieuDarcy: In that case, I believe the simplest option is to have it as a callable that takes in the data?

mattlevine22: This may not be the only place we would want to make choices based on data dtype. 

My preference would be to either skip it here entirely or step back and think about the more general version of this.

MatthieuDarcy: Maybe set it to the float32 default for the purpose of this PR and revisit later?

mattlevine22: fine by me

MatthieuDarcy: done

dynestyx/inference/configs/filter.py
```diff
@@ -134,6 +136,16 @@ class EnKFConfig(BaseFilterConfig):
...
             \(\sqrt{1 + \delta}\) before the update to prevent collapse.
             `None` disables inflation.
+        recorded_filtered_states_cov_jitter (float): Nonnegative \(\epsilon\) added to
+            the **recorded** filtered-state covariance as \(\epsilon I\).
+            This only affects the covariance when converted to a `MultivariateNormal` or `LowRankMultivariateNormal`
+            distribution (notably those returned in `ConditionedResult.dists`); it never
+            enters the EnKF update, the filter recursion, or the marginal
+            likelihood.
+            When `n_particles - 1 < state_dim`, the ensemble covariance is singular,
+            this regularization is necessary to give the recorded distributions a well-defined density (sampling will work nonethelss).
```
mattlevine22: nonetheless typo

dynestyx/inference/configs/discretizer.py
```diff
@@ -4,29 +4,22 @@
...
-        )
+    _validate_nonnegative_float("covariance_jitter", covariance_jitter)
 
 
 def _validate_jitter_scale(jitter_scale: float) -> None:
-    if not math.isfinite(jitter_scale) or jitter_scale < 0.0:
-        raise ValueError(
-            f"jitter_scale must be a finite, nonnegative float, got {jitter_scale!r}."
-        )
+    _validate_nonnegative_float("jitter_scale", jitter_scale)
```
mattlevine22: Let's remove these functions entirely and use `_validate_nonnegative_float` directly inline as needed.

tests/test_filters.py
```diff
@@ -669,6 +672,83 @@ def test_cuthbert_enkf_sparse_h_matches_dense_h():
...
+        assert isinstance(d, dist.LowRankMultivariateNormal)
+        assert d.event_shape == (dynamics.state_dim,)
+        assert d.cov_factor.shape == (dynamics.state_dim, n_particles)
+        assert jnp.allclose(d.mean, ensemble[t].mean(axis=0), atol=1e-5)
+        assert jnp.isfinite(d.sample(jr.PRNGKey(t))).all()
+
+    # Rebuilt exactly as `run_discrete_filter` does it: with the config's own
```
mattlevine22: is this second chunk really necessary? The above seems to cover things sufficiently.

tests/test_filters.py
mattlevine22: Thanks! I added a comment to simplify `test_cuthbert_enkf_filtered_dists_are_low_rank_and_samplable`

tests/test_distribution_utils.py
```diff
@@ -65,6 +67,118 @@ def test_cholesky_state_sequence_to_dists_gaussian():
...
+        ensemble=ensemble, mean=ensemble.mean(axis=-2), chol_cov=chol_cov
+    )
+
+    dists = _cholesky_state_sequence_to_dists(states, particle_mode=False)
+
+    assert isinstance(dists[0], dist.MultivariateNormal)
+    assert jnp.isfinite(dists[0].log_prob(dists[0].mean))
+
+
+def test_covariance_jitter_shifts_only_the_covariance_diagonal():
```
mattlevine22, line 121: This seems kinda excessive for just a jitter? What do you think? I would be okay omitting this, but if you picture this test protecting us from doing something stupid later, I'm fine with keeping it.

dynestyx/inference/configs/discretizer.py
```diff
@@ -4,29 +4,22 @@
...
 import diffrax as dfx
 
 from dynestyx.inference.configs.simulator import (
     ODESimulatorConfig,
     SDESimulatorConfig,
 )
+from dynestyx.utils import _validate_nonnegative_float
 
 
 def _validate_covariance_jitter(covariance_jitter: float) -> None:
```
mattlevine22: remove?

mattlevine22 (changes requested):
Looks great! Some small remaining (rather nit-picky, but easy) suggestions.

- Made some suggestions to simplify the tests, but if you want to keep those sections, that is fine.
- Pointed out some very thin functions that can be dropped.

dynestyx/inference/configs/filter.py
MatthieuDarcy: that's how you know this was an organically made commit. fixed

dynestyx/inference/configs/discretizer.py
MatthieuDarcy: agreed

tests/test_filters.py
MatthieuDarcy: Agree, I removed it and just added a check in the above that we get valid log prob.

tests/test_distribution_utils.py
MatthieuDarcy: I was being cautious because this is the only check that the jitter produces the right distribution. If we change `_cholesky_state_sequence_to_dists` in some way we might return a wrong distribution withough knowing (all others check only that`sample` and `log_prob` give finite values, no check of correctness). 

I don't know if I'm being too conservative (especially since we already run a lot of tests), happy to remove/trim it.

tests/test_filters.py
MatthieuDarcy: Thanks Matt, I simplified as per your suggestion.

mattlevine22 (approved):
This is great, thank you for making this happen!

Noting that another related improvement for EnKF with high-dim systems is on its way for the cuthbert backend (https://github.com/state-space-models/cuthbert/issues/277).

PR #359, opened 2026-09-11
Expose discretization, observation marginalization, and local drift linearization

ma9o:
**Summary**

Related: [issue #360](https://github.com/BasisResearch/dynestyx/issues/360).

Expose model operations for inference loops outside dynestyx:

- `discretize_dynamics` and `discretize_state_evolution` expose the configuration-driven conversion used by `Discretizer`, returning existing model and state-evolution types.
- `masked_observation_log_prob` evaluates the observed marginal of an observation distribution through the existing missing-observation scorer.
- `linearize_drift` constructs an `AffineDrift` by linearizing in state at fixed control and time. `LocalLinearizationConfig` uses the same helper.

**Notes**

Also fixes batch-axis handling in the shared Gaussian marginal calculation.


dynestyx/__init__.py
```diff
@@ -5,7 +5,11 @@
 __version__ = version("dynestyx")
 
 from dynestyx.api import log_prob, simulate
-from dynestyx.discretizers import Discretizer
+from dynestyx.discretizers import (
+    Discretizer,
+    discretize_dynamics,
```
ma9o, line 10: `discretize_dynamics` for whole models, `discretize_state_evolution` for callers that keep their own initial and observation models. Existing test files already imported the private one.

dynestyx/models/drifts.py
```diff
@@ -130,6 +131,35 @@ def __call__(
         return out
 
 
+def linearize_drift(
```
ma9o, line 134: Returns the existing `AffineDrift`, linearized in state with control and time fixed. Passing `state_evolution.total_drift` includes the potential-gradient contribution. The supplied control is incorporated into the drift, so `B=None`.

dynestyx/discretization/gaussian.py
```diff
@@ -43,18 +44,17 @@ def _local_linearization_moments(
     covariance_jitter: float,
 ) -> _Moments:
     h = _positive_interval(t_now, t_next)
-    f0 = cte.total_drift(x=x, u=u, t=t_now)
-    J = jax.jacfwd(lambda z: cte.total_drift(x=z, u=u, t=t_now))(x)
+    drift = linearize_drift(cte.total_drift, x=x, u=u, t=t_now)
```
ma9o, line 47: Uses the same drift helper. For constant additive diffusion, callers can compose it with `ExactAffineConfig` and `params_at` to recover the local transition parameters, using the same covariance jitter.

dynestyx/observation_missingness.py
```diff
@@ -652,6 +655,61 @@ def probe_observation_distribution_contract(
 
 
 def masked_observation_log_prob(
```
ma9o, line 657: Accepts an observation distribution, including one returned by a callable model. It shares the existing marginalization code with prepared scorers and the HMM, preserving distribution batch axes. Prepared paths retain their family and event-shape checks.

docs/api_reference/public/models/core/discrete_time_state_evolution.md
```diff
@@ -5,3 +5,27 @@
...
+
+
+def score_from(previous_state):
+    transition = dynamics.state_evolution(
+        x=previous_state, u=control, t_now=t_now, t_next=t_next
+    )
+    return jax.vmap(transition.log_prob, out_axes=-1)(states)
+
+
+pairwise_log_prob = jax.vmap(score_from, out_axes=-2)(previous_states)
```
ma9o, line 23: `vmap` composes the existing transition distributions. The inner and outer `out_axes` preserve `(*log_prob_batch, n_previous, n_current)` for dense all-pairs scores.

mattlevine22:
Thanks @ma9o, these look very reasonable!

1. I like the creation / exposition of `discretize_dynamics` and `discretize_state_evolution`, as well as `linearize_drift`

2. I like the idea of exposing `masked_observation_log_prob`, but I am a bit confused about:
a. What is the motivation for a `log_prob_batch` here? Do we need it at all? Could a user vmap instead (perhaps not, as @DanWaxman has showed that it is hard to `vmap` over numpyro distributions)?
b. "Also fixes batch-axis handling in the shared Gaussian marginal calculation."
- what is being fixed exactly? Was there a bug in some existing workflow? Or is this a fix needed to support the new stuff you've built?
- If there is a bug that you've fixed, could you please point to an associated regression test (or post a test that will fail on our `main` branch)?

3. I asked codex to look for bugs, and it found an issue with batching under shared covariances. Again, not sure what your intended use case for the batched version of this function is, so let's nail that down first.

```python

def test_masked_observation_log_prob_particles_share_sensor_covariance():
    # Three particles predict different means but share sensor noise.
    means = jnp.array([[0.1, 0.2], [0.3, -0.4], [-0.2, 0.5]])
    R = jnp.array([[0.5, 0.1], [0.1, 0.3]])
    obs_dist = dist.MultivariateNormal(means, covariance_matrix=R)

    actual = dsx.masked_observation_log_prob(
        obs_dist,
        y=jnp.array([0.4, jnp.nan]),
        obs_mask=jnp.array([True, False]),
    )

    # Only the first sensor is observed: one marginal likelihood per particle.
    expected = dist.Normal(means[:, 0], jnp.sqrt(R[0, 0])).log_prob(0.4)
    assert actual.shape == (3,)
    assert jnp.allclose(actual, expected)
```

docs/api_reference/public/discretizers/discretizer.md
```diff
@@ -1,5 +1,45 @@
...
+```
+
+Sampling requires a transition with `sample`; scoring additionally requires
+`log_prob`. `EulerMaruyamaConfig` supplies both for its Gaussian approximation.
+`DiffraxSampleConfig` supplies sampling only.
+
+For dense all-pairs scores, [map over the transition distributions](../models/core/discrete_time_state_evolution.md#all-pairs-transition-scores).
+For missing observations, use [masked_observation_log_prob](../models/core/observation_model.md#masked_observation_log_prob).
+
+## Effect-handler form
```
DanWaxman: I'd like for this to be smoothed over a bit. In particular, we should explain that there is an effect handler form, and an effect-handler-free form, and tell the user why they would want to choose one or the other. Moreovoer, if the effect handler form is the primary internal way to write things, we should also put that first.

ma9o:
Thanks for the quick feedback @mattlevine22 ! 

- regarding `masked_observation_log_prob`, a user could well just vmap around it but making the API support batching directly would be conformant to the batch/event shape semantics. e.g. cd-dynamax [passes](https://github.com/hd-UQ/cd_dynamax/blob/e0e56a6e05e058071f2e54fb26a9cc9275774a5d/cd_dynamax/src/continuous_discrete_nonlinear_ssm/inference_dpf.py#L365-L372) the whole particle array into one `emission_distribution.log_prob`. My use case is running a sampler on data with missing observations, and I thought the observation-likelihood semantics in general make sense to belong to dynestyx. 
- the "fix" is more like an extension to preserve batch axes for the new addition. There is nothing wrong with the current code, my bad.
- good find on the shared covariance, I was on JAX 0.11.1 and looks like it fails on 0.8.2. Not sure if this PR alone is worth restricting the range for.

mattlevine22:
1. regarding `masked_observation_log_prob`...:
- The example you point to in CD-Dynamax has a batch over `particles`, but not over the distribution; that is, it evaluates the log prob of a fixed distribution against many candidate samples from it. This indeed matches your use case, but should involve a batch over `y` and `obs_mask`, not over the `obs_dist`.
- In both `_masked_multivariate_normal_log_prob` and `masked_observation_log_prob`, the jaxtyping indicates that `y` and `obs_mask` should not have a batch dimension; thus, I'm assuming your intention is for the batch dimension to only appear in the `obs_dist`.
- Hence my confusion---what is the use case for having a fixed single `y` but a batch of distributions managed by this function (I'm very open to it, I just want to be sure I'm understanding things correctly)? I can imagine maybe this is relevant when doing hierarchical/mixed effects models and or in `Predictive` scenarios....but I also thought we handled that elsewhere/at a higher level.

2. Re: the "fix" is more like an extension to preserve batch axes for the new addition. There is nothing wrong with the current code, my bad.
- Got it, thanks!

3. "good find on the shared covariance, I was on JAX 0.11.1 and looks like it fails on 0.8.2. Not sure if this PR alone is worth restricting the range for."
- ah interesting! I agree not worth restricting here.

ma9o:
> The example you point to in CD-Dynamax has a batch over particles, but not over the distribution

But the distributions batch [gets built internally](https://github.com/hd-UQ/cd_dynamax/blob/e0e56a6e05e058071f2e54fb26a9cc9275774a5d/cd_dynamax/src/continuous_discrete_nonlinear_ssm/cdnlssm_utils.py#L64-L71) anyway no? My use case is also scoring particles and this would make it explicit in the interface.

> I can imagine maybe this is relevant when doing hierarchical/mixed effects models and or in Predictive scenarios

Now that you mention this maybe it makes sense to also support the other permutations of batching for samples and distributions (batch|single, single|batch, batch|batch, single|single), perhaps as a followup. (I got Astra to try it out [here](https://github.com/BasisResearch/dynestyx/commit/06a3c29387a58f4f2bae31d8a367c1d4d63f4537))

> But I also thought we handled that elsewhere/at a higher level

Is that wrt batching or also missing-data handling? If the dynamax snippet just above is the right example, it seems like the consumers need to carry their own batching logic that could live in dynestyx.

mattlevine22:
@DanWaxman when you have a chance, could you take a look at the batching story above?

DanWaxman:
> @DanWaxman when you have a chance, could you take a look at the batching story above?

Yes, I'll take a look today or tomorrow (sorry @ma9o for the delay, it's ICLR week :))

ma9o:
No rush :)

dynestyx/observation_missingness.py
```diff
@@ -652,6 +654,61 @@ def probe_observation_distribution_contract(
...
```
DanWaxman: Is there a particular reason to keep the public and private one separate?

DanWaxman (changes requested):
This looks very reasonable! The batching story makes sense to me (e.g., in a particle filter, you are building $N$ distributions $p(y| \mu_n, \sigma_n)$, scored against one obseration $y$ --- i.e., the parameters are batched against one evaluation). 

I mostly have some feedback about the docs and the introduction of a new private function.

PR #361, opened 2026-09-13
Infer control dim for B or D in LTI

mattlevine22:
Previously, the LTI classes inferred the `control_dim` only using `B`; if `B=None` and `D=jnp.array(obs_dim, ctrl_dim)`, it would fail.

Fixed to look at `D`'s shape as well if `B` is None.


DanWaxman (approved):
LGTM (though, with the caveat that this doubles down on the 1-d states/control thing that @MatthieuDarcy was having issues in)

mattlevine22:
I think we will deal with that separately. This is just a simple thing that makes the current code better for now.

MatthieuDarcy (approved):
LGTM. 

For ease of use, I would consider raising an error if B and D have different dimensions. Something like 

```
if B is not None and D is not None and B.shape[-1] != D.shape[-1]:
    raise ValueError(
        f"B and D must share the control dimension; got B.shape={B.shape}, D.shape={D.shape}"
    )
```

but optional.

PR #371, opened 2026-09-20
Fix failing tests after NumPyro upgrade

### LINKED ISSUE #370: Fix Failing Tests From NumPyro Upgrade
NumPyro recently released a v0.22.0, which [fixes some default behavior of validation](https://github.com/pyro-ppl/numpyro/pull/2201) and (rightly) [makes some tests fail](https://github.com/BasisResearch/dynestyx/actions/runs/35521242702/job/106105589620).

DanWaxman:
Resolves #370. 
- Uses a shorter smoketest dataset for `data_conditioned_continuous_time_deterministic_l63` that avoids singular values.
- Uses a jitter in `test_cholesky_state_sequence_to_dists_ensemble_is_low_rank`
- Turns off validation for `test_covariance_jitter_shifts_only_the_covariance_diagonal` (the 0-jitter is on purpose)


tests/fixtures.py
```diff
@@ -523,7 +523,7 @@ def data_conditioned_continuous_time_deterministic_l63(request):
     # ---------------------------------------------------------
     # Generate observations at some times
     predict_times = _profiled_arange(
-        fast_stop=1.0, fast_step=0.1, science_stop=2.0, science_step=0.001
+        fast_stop=0.5, fast_step=0.1, science_stop=2.0, science_step=0.001
```
mattlevine22, line 526: I guess this is OK...the point is just to smoke the composition of all these functions (it is "bad" that we needed to use a smaller fast_stop though).

tests/test_distribution_utils.py
```diff
@@ -77,10 +78,13 @@ def test_cholesky_state_sequence_to_dists_ensemble_is_low_rank():
     the one every other EnKF test misses by using ``state_dim=2``.
     """
     t_len, n_particles, state_dim = 3, 4, 16
+    jitter = 1e-5
     ensemble = jr.normal(jr.PRNGKey(0), (t_len, n_particles, state_dim))
     states = SimpleNamespace(ensemble=ensemble)
 
-    dists = _cholesky_state_sequence_to_dists(states, particle_mode=False)
+    dists = _cholesky_state_sequence_to_dists(
```
mattlevine22, line 85: So I think this test previously was designed to check that we can get still samples from a jitter-free path (and that despite all the new "jitter" code, we indeed were remaining jitter-free in that case).

Should we stop allowing the jitter-free lifestyle altogether?

mattlevine22: not even sure if we need this test at this point

mattlevine22 (commented):
Thanks for doing this!
I'm happy to approve as is, but wrote some comments / questions to check on some stuff.

tests/test_distribution_utils.py
MatthieuDarcy: This path was supposed to do validate that `_cholesky_state_sequence_to_dists` gives the correct distribution (Low Rank with the correct covariance, dimensions etc...). For this purpose, using `jitter` non zero is fine imo. It's maybe a bit redundant with the `test_covariance_jitter_shifts_only_the_covariance_diagonal` test below, but I feel that having a test to validate a basic operation is good, what do you think? I can maybe look to simplify/unify some tests.

Re: allow jitter-free lifestyle, I am not a fan of putting an arbitrary floor to accuracy and I believe we should give users a knob to adjust this.

MatthieuDarcy (approved):
I left a comment on one of Matt's comment but otherwise I think this is good as is.

tests/fixtures.py
DanWaxman: Right... it's just because EKF blows up with the bigger step size, especially with 32-bit floats

tests/test_distribution_utils.py
DanWaxman: Yeah, they will generally have a knob but also have to turn validation off. Maybe we should make a PR to NumPyro making an argument to allow zero diagonal (though it does make things like logprobs bad)

MatthieuDarcy: I think that `LowRank` should probably always use a jitter (but it's hard to choose a good floor, do we assume float32 or float64?). The dense path on the other hand might legitimately want zero jitter and I think it's not super intuitive to treat them differently. So I think our compromise of have a single parameters for both, with the chance of failing, is tolerable (open to counter points). 

I think that maybe raising a warning if jitter is zero in the low rank case would not be a bad idea.

